PR #20, opened 2023-10-27
Adding Rule Set addresses and contents

blockiosaurus:


src/pages/token-auth-rules/mplx-rule-sets.md
```diff
@@ -0,0 +1,73 @@
+---
+title: Overview
+metaTitle: Token Auth Rules - Metaplex Rule Sets
+description: The pNFT Rule Sets that Metaplex Foundation manages.
+---
+The Metaplex Foundation manages two Rule Sets for pNFTs.
+
+## Metaplex Foundation Rule Set
+This rule set represents the best efforts of finding and blocking programs that do not enforce marketplace royalties.
```
febo, line 9: Can we add a note here to explain why it currently looks like the "Compatibility Rule Set"? E.g., at the moment there are no programs bypassing creator royalties, therefore the rule set does not block any programs.

src/pages/token-auth-rules/mplx-rule-sets.md
```diff
@@ -0,0 +1,73 @@
+---
+title: Overview
+metaTitle: Token Auth Rules - Metaplex Rule Sets
+description: The pNFT Rule Sets that Metaplex Foundation manages.
+---
+The Metaplex Foundation manages two Rule Sets for pNFTs.
```
febo, line 6: Is it worth adding a note to say how these get updated?

febo (commented):
Left a couple of comments.

febo (approved):

PR #22, opened 2023-12-07
clarify amounts for token guards with sugar

MarkSackerberg:
Users on discord were confused on how to supply the amount when using the sugar config.
This is an attempt to clarify it. If someone has a better Idea on how to clarify it - feel free to suggest something!

### DIFF
--- a/src/pages/candy-machine/guards/freeze-token-payment.md
+++ b/src/pages/candy-machine/guards/freeze-token-payment.md
@@ -116,7 +116,7 @@ Add this object into the guard section your config.json file:
 
 ```json
 "freezeTokenPayment" : {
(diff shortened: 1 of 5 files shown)


src/pages/candy-machine/guards/freeze-token-payment.md
```diff
@@ -116,7 +116,7 @@ Add this object into the guard section your config.json file:
 
 ```json
 "freezeTokenPayment" : {
-    "amount": number,
+    "amount": number number in basis points (e.g. 1000 for 1 Token that has 3 decimals),
```
lorisleiva, line 119: I think there's a double "number" now haha.

MarkSackerberg: (╯°□°)╯︵ ┻━┻

Thank you!

PR #23, opened 2023-12-08
Add token-2022 page

febo:
This PR adds details on how SPL Token-2022 works with Token Metadata.


blockiosaurus:
Nit: would it be possible to make the tables render with a consistent width across all of them? It seems column width is random.

blockiosaurus (approved):
Perfection!

PR #24, opened 2023-12-12
Update Sign Command Documentation

0xIchigo:
This PR aims to update Sugar's `sign` command documentation. The updated documentation makes users aware that the command relies on a standard gPA call to Metaplex's Token Program (pursuant to [PR #389](https://github.com/metaplex-foundation/sugar/pull/389)). The update emphasizes that such calls can be resource-intensive and might be blocked by some RPC providers

### DIFF
--- a/src/pages/candy-machine/sugar/commands/sign.md
+++ b/src/pages/candy-machine/sugar/commands/sign.md
@@ -17,3 +17,13 @@ And running with a specific keypair:
 ```
 sugar sign -k creator-keypair.json
 ```
+
+Note using `sugar sign` relies on a standard `getProgramAccounts` call to the Metaplex Token Metadata program (i.e., `metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s`). Ensure that the RPC node you use supports standard gPA calls to that address, as it can be pretty intensive. Developers can provide a custom RPC URL with the command:
+```
+sugar sign -r <RPC_URL>
+```
+
+Alternatively, NFTs can be signed one at a time using the command:
+```
+sugar sign -m <MINT_ADDRESS>
+```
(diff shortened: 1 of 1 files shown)


src/pages/candy-machine/sugar/commands/sign.md
```diff
@@ -16,4 +16,13 @@ And running with a specific keypair:
 
 ```
 sugar sign -k creator-keypair.json
+```
+
+Developers can provide a custom RPC URL with the command:
+```
+sugar sign -r <RPC_URL>
+```
+Note using `sugar sign` relies on a standard `getProgramAccounts` call to the Metaplex Token Metadata program (i.e., `metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s`). This is very inefficient. The recommended solution is to sign NFTs individually using the command:
```
blockiosaurus, line 25: I think my preference would be towards more neutral language here. Otherwise this is great!

`Note using `sugar sign` relies on an inefficient `getProgramAccounts` call on the Metaplex Token Metadata program (i.e., `metaqbxxUerdq28cj1RbAWkYQm3ybzjb6a8bt518x1s`). The recommended solution is to sign NFTs individually using the command:`

0xIchigo: I've updated the PR to make its language more neutral with the following [commit](https://github.com/metaplex-foundation/developer-hub/pull/24/commits/d4054081a0626045bddd45f9d4c3fd6771e0811a). Let me know what you think!

blockiosaurus: Nit: can you say "on the Token Metadata program" instead of "to the Token Metadata program" since gPA is not an instruction in the Token Metadata program?

0xIchigo: Sorry, just seeing this now - I've updated the wording @blockiosaurus

blockiosaurus (approved):

PR #42, opened 2024-02-20
Add legacy docs

MarkSackerberg:
Copying over the docs from docs.metaplex.com before it is gone for good.


MarkSackerberg:
should be fixed @tonyboylehub

PR #55, opened 2024-03-18
Align Fee page

MarkSackerberg:
There were no changes to fees, but the page was cluttered with history that should not be required anymore.

This should make it easier to understand the current fee structure.


MarkSackerberg:
obsolete with the core merge now

PR #68, opened 2024-03-28
Make github header link open in new tab

wooddyy-sol:
Found it inconvenient for the github link in the header to redirect within the same tab, as I was usually reading docs and wanting to goto github to look at some source code without losing my place.


MarkSackerberg (approved):

MarkSackerberg:
Thank you!
This change makes sense.
Merged!

PR #77, opened 2024-04-08
docs(token2022PaymentGuard): fix minting code block

MohammedAlabd:
when minting the user should specify that it is a `token2022Payment` and not a `tokenPayment` otherwise it will pass the token22 ID and it will throw an error


MarkSackerberg (approved):

MarkSackerberg:
you are totally correct here. Thank you very much!

PR #93, opened 2024-04-30
Add new core cm guards

MarkSackerberg:


src/pages/core-candy-machine/guards/nft-mint-limit.md
```diff
@@ -0,0 +1,139 @@
+---
+title: "NFT Mint Limit"
+metaTitle: "Core Candy Machine Guards - NFT Mint Limit"
+description: "The NFT Mint Limit guard restricts minting to holders of a specified NFT collection and limits the amount of mints that can be done for a provided NFT."
+---
+
+## Overview
+
+The NFT Mint Limit guard restricts minting to holders of a specified NFT collection and limits the amount of mints that can be done for a provided NFT. It can be considered as a combinatino of the [NFT Gate](/core-candy-machine/guards/nft-gate) and [Mint Limit](/core-candy-machine/guards/mint-limit) Guard, based on NFT Addresses instead of wallets.
```
blockiosaurus, line 9: Typo: "combinatino"

blockiosaurus, line 9: I think we should specify somewhere that this guard is for Token Metadata assets, not Core assets.

MarkSackerberg: you are right, I've added "Token Metadata" in there. Will also check the other "nft" related guards for this. Users will not be aware that NFT = TM and Asset = Core.

tonyboylehub (approved):

PR #103, opened 2024-05-14
Adding Plexpert

blockiosaurus:


tonyboylehub (approved):

MarkSackerberg:
When plexpert is loaded the styling changes from
before
to
after

MarkSackerberg:
Asking on how to create a NFT using Umi the script returned is not successful.
E.g. this `umi.create` does not seem correct and should be `await createV1(...)` according to our docs instead 
grafik

Did you feed all the core docs into it already?

MarkSackerberg (changes requested):
Even when seeing this as test version which is not always correct we should fix:

-  the styling.
- When the Chat window is open the menu / all links become unusable

blockiosaurus:
Alas, it's a simple embed so not sure how much of the styling issues I can control. But I'll check, thank you for the feedback!

PR #107, opened 2024-05-18
patch: remove creators details in master edition docs

jim4067:
I think they add confusion to the example, given that these details can only be added using the Royalty plugin.


tonyboylehub:
I noticed these and fixed them in a different PR this morning. This will be fixed a bit later in a different PR.

PR #118, opened 2024-05-31
Cleaning up the grid by splitting it into sections.

blockiosaurus:


MarkSackerberg:
I really like the Idea!
The 404 error Page is broken now though:
image
The normal "Programs and Tools" page, too
image

tonyboylehub:
I had to make mulitple grids for this very reason to stop things getting messed up as all these parts where connected to each other.

blockiosaurus:
Ah no it was not. I think moving the className broke everything so let me try and fix. Thanks for finding these!

tonyboylehub (approved):

PR #125, opened 2024-06-06
added working with rust page to home page

tonyboylehub:


blockiosaurus:
This is great! Thank you! Can you copy and paste a subset of this page to the mpl-core as well? I think that's getting most of the hits for people asking for Rust support, but this is super useful to point to in general.

tonyboylehub:
Added a smaller version of this with some Core specific code such as `cargo add`.

src/pages/core/getting-started/rust.md
```diff
@@ -4,4 +4,100 @@ metaTitle: Core - Getting Started - Rust
...
+You can add the `mpl-core` crate to your Rust project by installing with `cargo`.
+
+```shell
+cargo add mpl-core
+```
+
+Alternatively you can paste the version of `mpl-core` crate you wish to use straight into your Cargo.toml file under `[dependencies]`.
+
+```toml
+mpl-core = "0.6.1"
```
blockiosaurus, line 24: If you exclude the version here it should just add the latest, which I think is what we want. Just so we don't have to always update this.

tonyboylehub: sweet, will replace with x.x.x

blockiosaurus (approved):

PR #136, opened 2024-06-25
candy machine docs changes for cu

jim4067:
With Core, I don't think requestion `800_000` compute units is reasonable. 

Remove the paragraph from the basic minting section and introduced it in minting with guards

Reduce initial requested CUs to 300k

Here is the [transaction](https://explorer.solana.com/tx/4nmF8PgkLKFdkJfTmk5yqoRDTFZPtFf87aaguy73cKBVojadvcn3CdA7AKAKXP9kZe83btqpSMLKn8g7BtQmiwvd?cluster=devnet) for a basic mint from a core CM using the sol guard using less that 80k CUs

### DIFF
--- a/src/pages/core-candy-machine/mint.md
+++ b/src/pages/core-candy-machine/mint.md
@@ -106,9 +106,9 @@ const coreCollection = publicKey("22222222222222222222222222222222");
 const asset = generateSigner(umi);
 
 await mintV1(umi, {
-	candyMachine: candyMachineId,
-	asset,
-	collection: coreCollection,
+  candyMachine: candyMachineId,
+  asset,
+  collection: coreCollection,
 }).sendAndConfirm(umi);
 
 ```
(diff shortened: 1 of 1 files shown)


danenbm (changes requested):
Thank you for the PR!  LGTM except could you please switch back to 2 space indent for the example code?

jim4067:
updated @danenbm

danenbm:
> updated @danenbm

Is it just me?  I'm still seeing some tabs that render in here as bigger than 2 spaces.  Would like to convert them to spaces (we normally have our IDEs do that for us):
```
await mintV1(umi, {
	candyMachine: candyMachineId,
	asset,
	collection: coreCollection,
}).sendAndConfirm(umi);
```

jim4067:
You are correct. Pushed the update

Weirdly prettier was telling me that it was using 2 spaces

danenbm (approved):
Awesome ty!

MarkSackerberg:
Thank you! Merged!

PR #138, opened 2024-06-26
Tony/serialization and deserialization

L0STE:
Added the deserialization doc


src/pages/core/deserialization.md
```diff
@@ -0,0 +1,158 @@
...
+println!("collection_V1: {:?}", asset_v1);
+```
+
+{% /dialect %}
+
+{% /dialect-switcher %}
+
+### Deserializing Plugins
+
+Even if the plugin data is saved directly on the Asset account, because of the size of it we should deserialize and save just the singular plugin data
```
danenbm, line 125: I think this wording should be changed, because the plugin data is stored on the asset account, and it is included in the Asset type.  But I think this wording will probably change based on my initial comment anyways.

src/pages/core/deserialization.md
```diff
@@ -0,0 +1,158 @@
...
+title: Deserialization
+metaTitle: Deserialization
+description: Learn about the deserialization of Asset accounts, Collection accounts and plugins.
+---
+
+Digital assets on Core are composed of exactly **one on-chain account** that has all the relevant data inside of it.
+
+That means that if we want to read that data we need to learn how to deserialize it.
+
+### Deserializing Assets
```
danenbm, line 11: Should we also add information about deserializing in Anchor in this guide?  Or save it for another guide? 
 i.e. suggesting that `Account<'info, BaseAssetV1>` can be used in Anchor (if you've tested it can be done that way).

src/pages/core/deserialization.md
danenbm, line 1: Page won't build.  I think you need to run `pnpm run build` and fix any issues

danenbm (changes requested):

src/pages/core/deserialization.md
tonyboylehub: `Account<'info, BaseAssetV1>` isn't usable for me directly with Anchor account decoding. There's something in there blocking it still but looks like everything has Anchor Serialization and Deserialization on it in the children of the struct but it's still complaining that bot Anchor Serialization and Deserialization are not satisfied.

danenbm: Yeah per our offline discussion, we need to implement additional stuff to allow deserialization via Account struct.  I would say don't let it block this guide, but we will see if we can get it fixed soon.

danenbm (dismissed):

tonyboylehub (changes requested):

tonyboylehub (approved):

PR #139, opened 2024-06-27
fixed collection

tonyboylehub:


src/pages/core/create-asset.md
```diff
@@ -230,16 +230,32 @@ MPL Core Assets can be created straight into a collection if your MPL Core Colle
...
 await createCollection(umi, {
   collection: collectionSigner,
   name: 'My Collection',
   uri: 'https://example.com/my-collection.json',
 }).sendAndConfirm(umi)
 
+// fetch the collection
+// if the createCollection transaction succeeds but
+// the fetch produces 'accountNotFound' then you will need
+// to sleep before executing the fetch.
```
danenbm, line 254: Or use finalized right?  This comment could be here but if it is meant to be redundant to the first one might be clearer to just remove it, or put the first comment over the fetch?

tonyboylehub: yeah you are right. Fixed.

danenbm (approved):

PR #143, opened 2024-07-05
feat: use pda address in token payment code examples

jim4067:


MarkSackerberg (approved):
Hey,
thanks for the suggestion!
I feel like it can make it easier to use for some people.
--> Merged!

PR #150, opened 2024-07-22
AppData serialization/deserialization + read/write

tonyboylehub:


src/pages/core/external-plugins/app-data.md
```diff
@@ -248,3 +349,178 @@ let write_to_app_data_plugin_ix = WriteExternalPluginAdapterDataV1CpiBuilder::ne
...
+The first step to deserializing the data stored in an AppData plugin is to fetch the raw data and check the schema field which dictates the format in which the data is stored before serialization.
+
+{% dialect-switcher title="Fetching AppData Raw Data" %}
+{% dialect title="JavaScript" id="js" %}
+
+```ts
+const assetId = publicKey('11111111111111111111111111111111')
+const dataAuthority = publicKey('33333333333333333333333333333333')
+
+const asset = await fetchAsset(umi, assetId)
```
danenbm, line 368: So for Rust client we recommend they only deserialize the parts they need.  But for JS don't we give them everything?  Doesn't `fetchAsset` deserialize the asset and all the internal and external plugins for them, plus parse the `AppData's` data?

I think the flow is `fetchAssetV1` -> `deserializeAssetV1` -> `getAssetV1AccountDataSerializer()` ->  `externalRegistryRecordsToExternalPluginAdapterList` -> `appDataFromBase` -> `parseExternalPluginAdapterData`.

So for typescript wouldn't we show just using fetchAsset and what they would get as a result which I think would be the data too?  Or am I missing something?

danenbm (changes requested):

src/pages/core/external-plugins/app-data.md
```diff
@@ -248,3 +349,178 @@ let write_to_app_data_plugin_ix = WriteExternalPluginAdapterDataV1CpiBuilder::ne
...
+
+{% /dialect %}
+
+{% dialect title="Rust" id="rust" %}
+
+```rust
+
+let plugin_authority = ctx.accounts.authority.key();
+
+let asset = BaseAssetV1::from_bytes(&data).unwrap();
```
danenbm, line 395: I wonder if for Rust if we should also show the second approach using `Asset::from_bytes()` and then using the `AppDataWithData::data_offset` and `AppDataWithData::data_len` to calculate the slice and then parse it from there?  Example of this would be in `create_with_external_plugin.rs` where it says "Second, get app data offset and length from a full `Asset` deserialization".

On the other hand maybe we don't want to showcase the full-on `Asset` and `Collection` as much.  But they are there and probably viable for any non-program Rust user (I think they work in programs now too as long as assets don't have too many plugins (After the stack optimizations it worked fine for me in the IslandMon Anchor game with a big `Attributes` plugin).

src/pages/core/external-plugins/app-data.md
```diff
@@ -248,3 +349,178 @@ let write_to_app_data_plugin_ix = WriteExternalPluginAdapterDataV1CpiBuilder::ne
...
+println!("{:?}", my_data);
+```
+
+{% /dialect %}
+
+{% /dialect-switcher %}
+
+#### Deserialze Binary Schema
+
+Because the **Binary** schema is arbitory data then deserialization will be dependent on the serialization you used.
```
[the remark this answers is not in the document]
tonyboylehub: These I don't and I add it to the code example itself as a comment is because I didn't want to disclose Rust stuff in the general docs the same way I don't disclose JS stuff in the general docs.

Unless you can think of a better way to do it? Otherwise I'll have to describe all the JS and rust packages in the made code rather than their switchable code blocks.

src/pages/core/external-plugins/app-data.md
```diff
@@ -248,3 +349,178 @@ let write_to_app_data_plugin_ix = WriteExternalPluginAdapterDataV1CpiBuilder::ne
...
+
+{% dialect-switcher title="Deserializing MsgPack" %}
+{% dialect title="JavaScript" id="js" %}
+```js
+As the binary data is arbitory you will need to include your own deserializers to
+parse the data into a usable format your app/website will understand.
```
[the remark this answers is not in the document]
tonyboylehub: Fix in future commit for switcher Title.

This was for JS, we don't have bincode in JS that I know of. How would you want to show Javascript deserialization and serialization of bytes? For now I just left it open ended as people wanting using binary serialization and deserialization I would assume would have their methods already chosen.

In the serialization example I just had a `Uinit8Array([1,2,3,4,5])` that was stored to data.

src/pages/core/external-plugins/app-data.md
tonyboylehub: Full JS example for reference.

```ts
const assetId = publicKey('11111111111111111111111111111111')
const dataAuthority = publicKey('33333333333333333333333333333333')

const asset = await fetchAsset(umi, assetId)

let appDataPlugin = asset.appDatas?.filter(
  (appData) => (appData.authority.address = dataAuthority)
)

let data
let schema

// Check if `AppData` plugin with the given authority exists
if (appDataPlugin && appDataPlugin.length > 0) {
  // Save plugin data to `data`
  data = appDataPlugin[0].data

  // Save plugin schema to `schema`
  schema = appDataPlugin[0].schema
}
```

> So for JS since we typically don't have the same memory constraints as a Rust program user, wouldn't we show just using fetchAsset and what they would get as a result which I think would be the data too? Or am I missing something?

`FetchAsset()` does return a fully deserialized asset but this example shows you how to grab a specific `AppData` plugin on the asset like we do in the Rust version. Maybe not needed?

danenbm: Yeah I believe its already deserialized, so then maybe you don't need to show that in JS because you already have it, unless its binary and you wanted to show using something extra like bincode.

src/pages/core/external-plugins/app-data.md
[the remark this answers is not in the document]
danenbm: Would it be beneficial to show bincode in JS the way it is shown in Rust?  Or if you are sticking with the array example you could show just like an assert or log or something just showing accessing the Uint8Array after its gotten?

This could also be handled based on what the JS deserialization examples look like based on the other comment about how `fetchAsset` already deserializes the data.

src/pages/core/external-plugins/app-data.md
```diff
@@ -248,3 +349,176 @@ let write_to_app_data_plugin_ix = WriteExternalPluginAdapterDataV1CpiBuilder::ne
...
+
+#### Deserialize Binary Schema
+
+Because the **Binary** schema is arbitrary data then deserialization will be dependent on the serialization you used.
+
+{% dialect-switcher title="Deserializing Binary" %}
+{% dialect title="JavaScript" id="js" %}
+```js
+As the binary data is arbitrary you will need to include your own deserializer to
+parse the data into a usable format your app/website will understand.
```
danenbm, line 501: Could you make these comments `//` so they match the other JS examples above?

danenbm (approved):

danenbm (approved):

PR #157, opened 2024-07-24
on-chain to onchain

MarkSackerberg:


src/pages/core-candy-machine/sugar/index.md
tonyboylehub, line 1: should this one be changed as it's referencing an external program's interface that wouldn't have been corrected. I'm ok with this either way, just wanted to bring it up.

MarkSackerberg: Technically you are right. I changed it nevertheless because of the SEO / Google perspective.

tonyboylehub (approved):

PR #161, opened 2024-07-26
add mpl-hybrid

MarkSackerberg:


L0STE (changes requested):
Was going through the updated official-link page and saw that the Candy Guard is linked to the Auction House Docs. I don't know if it's right. Other than this, everything else is in order

L0STE (approved):

PR #173, opened 2024-08-06
added sitemap generator

tonyboylehub:
This generates a working site map at  `/sitemap.xml` from the root domain using a edge function to generate a sitemap on demand. This works better than the "manually generated" one I origiainlly build.

/sitemap.xml will not work locally as it's build to Vercels build deployment and folder structure.


tonyboylehub:
LFG! Glad that's over with!

Comment from `sitemap.xml.js`

```
// Turns out Vercel builds the site into a different directory and file structure than running `next build` locally
// So we need to read the routes-manifest.json file from a different location than the .next directory that is locally
// built. The /sitemap.xml route will not work on local development, but it will work on Vercel deployment.
```

Turns out local builds also put the pages into `staticRoutes` array, where as Vercel puts the pages into `dataRoutes` array for what ever reason. Was running around chasing the wrong thing thinking the builds would be the same.

L0STE (approved):

PR #174, opened 2024-08-06
Reworked and Merged Web3Js Adapter and Differences Page

L0STE:


src/components/products/umi/index.js
```diff
@@ -54,7 +54,6 @@ export const umi = {
             { title: 'Serializers', href: '/umi/serializers' },
             { title: 'Storage', href: '/umi/storage' },
             { title: 'Transactions', href: '/umi/transactions' },
-            { title: 'Web3Js Adapters', href: '/umi/web3js-adapters' },
```
MarkSackerberg, line 57: This will change the URI. Therefore all (external) links pointing to those, e.g. on stackoverflow will be 404.

I feel like we should not do this. It was a huge pain when we migrated from docs.metaplex.com. At least we should have a solution to forward to the new page. Not sure how to best do it from SEO perspective. @tonyboylehub ?

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+When creating `Umi`, we wanted to move away from the class-based types defined in `@solana/web3.js` and instead use a more functional approach by relying only on TypeScript types. This unfortunately means that, although having the same or similar import names, not all types from `@solana/web3.js` are compatible with the ones provided by `Umi` and vice versa.
+
+To help with this issue, `Umi` provides a set of adapters that allows to parse types to and from their `Web3.js` counterparts and they can be found in the [`@metaplex-foundation/umi-web3js-adapters`](https://www.npmjs.com/package/@metaplex-foundation/umi-web3js-adapters) package.
+
+## Required Package and Imports
+
+To install the `umi-web3js-adapters` package and be abel to access to a bunch of helper methods to convert to and from Web3.js types, use the following commands:
+
+```
+npm i @metaplex-foundation/umi-web3js-adapters
```
MarkSackerberg, line 18: If you have @metaplex-foundation/umi installed you do not need this package.
Quite many people feel (for whatever reason) adding more packages as "not good" (even though they are dependent on each other, or even included like here either way)

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+  // Keypairs
+  fromWeb3JsKeypair, toWeb3JsKeypair,
+  // Publickey
+  fromWeb3JsPublicKey, toWeb3JsPublicKey,
+  // Legacy Transactions
+  fromWeb3JsLegacyTransaction, toWeb3JsLegacyTransaction,
+  // Versioned Transactions
+  fromWeb3JsTransaction, toWeb3JsTransaction, 
+  // Messages
+  fromWeb3JsMessage, toWeb3JsMessage, toWeb3JsMessageFromInput
```
MarkSackerberg, line 36: How did you order those? I would go either thematic (Instruction, LegacyTransaction v0Transaction, Message, Pubkey) or alphabetical?

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+} from '@metaplex-foundation/umi-web3js-adapters';
+```
+
+## Publickeys
+
+Generating public keys might seem similar at first sight, but there are some subtle differences between the packages. Web3Js uses a capital `P` and requires `new`, while the Umi version uses a lowercase `p`.
```
MarkSackerberg, line 42: we should stick to one notation. You have Publickeys in the header, public keys in the text

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+// Generate a new Publickey
+const umiPublicKey = publicKey("11111111111111111111111111111111");
+
+// Convert it using the UmiWeb3jsAdapters Package
+const web3jsPublickey = toWeb3JsPublicKey(umiPublicKey);
+```
+
+## Keypairs
+
+Generating keypairs is where the difference from Web3Js and Umi increase. For Web3Js we can just use the `Keypair.generate()` function, for Umi we need a Umi instance before.
```
MarkSackerberg, line 88: Maybe add that this umi instance is used for most (if not all useful) applications in umi?

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+// Generate a new keypair
+const umiKeypair = generateSigner(umi)
+
+// Convert it using the UmiWeb3jsAdapters Package
+const web3jsKeypair = toWeb3JsKeypair(umiKeypair);
+```
+
+## Instructions
+
+When creating instructions, the main difference is that with Umi, you need to create a Umi instance first (like the `Keyapair`) and using the `getInstructions()` you receive an array of instructions rather than a single one.
```
MarkSackerberg, line 148:
```suggestion
When creating instructions, the main difference is that with Umi, you need to create a Umi instance first (like the `Keypair`) and using the `getInstructions()` you receive an array of instructions rather than a single one.
```

MarkSackerberg, line 148: I'd mention that for many use cases, where only umi is used, handling single instructions is not needed but can be abstracted away using helpers and transaction builders.
Most cases where I see people actually "manually" handling instructions it's more of a foodgun for them then anything else

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+// get instructions will give you an Array of instructions
+const umiInstructions = transfer(umi, {...TransferParams}).getInstructions();
+```
+
+### Web3Js
+```ts
+import { SystemProgram } from '@solana/web3.js';
+
+// Create a new instruction (like a lamport transfer)
+const web3jsInstruction = SystemProgram.transfer({...TransferParams})
```
MarkSackerberg, line 168: This is a sol transfer. Above you show a MPL Asset transfer. Wouldn't it make sense to stay in the same "category"

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+// Convert it using the UmiWeb3jsAdapters Package
+const web3jsInstruction = umiInstruction.map(toWeb3JsInstruction);
+```
+
+## Transactions
+
+The Solana runtime supports two transaction versions:
+- Legacy Transaction: older transaction format with no additional benefit
+- 0 / Versioned Transaction: added support for Address Lookup Tables
```
MarkSackerberg, line 208:
```suggestion
- Legacy Transaction: Older transaction format with no additional benefit
- 0 / Versioned Transaction: Added support for Address Lookup Tables
```

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+**Note**: if you're not familiar with the concept of Versioned Transactions, read more about it [here](https://solana.com/it/docs/advanced/versions)
```
MarkSackerberg, line 210:
```suggestion
**Note**: if you're not familiar with the concept of Versioned Transactions, read more about it [here](https://solana.com/en/docs/advanced/versions)
```

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+import { createUmi } from '@metaplex-foundation/umi-bundle-defaults'
+import { transfer, mplCore } from '@metaplex-foundation/mpl-core'
+// Generate a new Umi instance
+const umi = createUmi('https://api.devnet.solana.com').use(mplCore())
+
+// Create a new Umi Legacy Transaction
+const umiTransaction = transfer(umi, {...TransferParams}).useLegacyVersion();
+
+// Create a new Umi Versioned Transaction
+const blockhash = await umi.rpc.getLatestBlockhash()
```
MarkSackerberg, line 225: I would move it down below `getInstructions` otherwise it might look like you need the blockhash for the instruction creation.

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+const instructions = transfer(umi, {...TransferParams}).getInstructions()
```
MarkSackerberg, line 227: Without `.getInstructions()` it would be a transaction builder already that has what you set below automatically. No need to be that verbose.
`transfer(umi, {...TransferParams}).build(umi)` would be type `Transaction` already and by default v0 with umi.identity as payer.

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+const umiVersionedTransaction = umi.transactions.create({
+  version: 0,
+  payer: frontEndSigner.publicKey,
```
MarkSackerberg, line 231: I feel like introducing a `frontEndSigner` here is confusing.

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+### Web3Js
+```ts
+import { Transaction, VersionedTransaction, TransactionMessage, Connection, clusterApiUrl, SystemProgram } from '@solana/web3.js';
+
+// Create a new Web3Js Legacy Transaction
+const web3jsTransaction = new Transaction().add(SystemProgram.transfer({...TransferParams}));
+
+// Create a new Web3Js Versioned Transaction
+const connection = new Connection(clusterApiUrl("devnet"));
+const minRent = await connection.getMinimumBalanceForRentExemption(0);
```
MarkSackerberg, line 246: you do not use this, do you?

MarkSackerberg (changes requested):

src/components/products/umi/index.js
MarkSackerberg: https://nextjs.org/docs/pages/api-reference/next-config-js/redirects

L0STE: Probably we can "keep" the old docs up but not let them "viewable" from the index? 

If I leave there the page and delete it just from the index, it will still work from an external link. In that space we can keep the "old page" or duplicate the new page to both url

src/pages/umi/web3js-differences-and-adapters.md
L0STE: Oh nice, I thought it was mandatory to install it!

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+The `@solana/web3.js` library is currently widely used in the Solana ecosystem and defines its own types for `Publickeys`, `Transactions`, `Instructions`, etc. 
+
+When creating `Umi`, we wanted to move away from the class-based types defined in `@solana/web3.js` and instead use a more functional approach by relying only on TypeScript types. This unfortunately means that, although having the same or similar import names, not all types from `@solana/web3.js` are compatible with the ones provided by `Umi` and vice versa.
+
+To help with this issue, `Umi` provides a set of adapters that allows to parse types to and from their `Web3.js` counterparts and they can be found in the [`@metaplex-foundation/umi-web3js-adapters`](https://www.npmjs.com/package/@metaplex-foundation/umi-web3js-adapters) package.
+
+## Required Package and Imports
+
+To install the `umi-web3js-adapters` package and be abel to access to a bunch of helper methods to convert to and from Web3.js types, use the following commands:
```
MarkSackerberg, line 15:
```suggestion
To install the `umi-web3js-adapters` package and be able to access to a bunch of helper methods to convert to and from Web3.js types, use the following commands:
```

src/components/products/umi/index.js
tonyboylehub: urls should not be touched unless redirected.

We are not keeping duplicate content or pages.

src/pages/umi/web3js-differences-and-adapters.md
tonyboylehub: Do we really need this? This isn't a "guide" per say more a functional referece. All the single imports should be viewable from the example code.

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+
+Next, let's look into how to use the adapters.
+
+### From Web3Js to Umi
+```ts
+import { PublicKey } from '@solana/web3.js';
+import { fromWeb3JsPublicKey } from '@metaplex-foundation/umi-web3js-adapters';
+
+// Generate a new Publickey
+const web3jsPublickey = new PublicKey("<1111111111111111111111111111111111111111>");
```
tonyboylehub, line 68: < > still inserted

src/pages/umi/web3js-differences-and-adapters.md
```diff
@@ -0,0 +1,377 @@
...
+// Generate a new Umi Publickey
+const umiPublicKey = publicKey("11111111111111111111111111111111");
+```
+
+### Web3Js
+```ts
+import { PublicKey } from '@solana/web3.js';
+
+// Generate a new Web3Js Publickey
+const web3jsPublickey = new PublicKey("<1111111111111111111111111111111111111111>");
```
tonyboylehub, line 57: < > still inserted

src/components/products/umi/index.js
L0STE: How does the redirect work?

MarkSackerberg: Multiple options. I did some testing and i feel like easiest would be to use the middleware. To do so:

1. add a middleware.js file into the `src` folder
2. add some code like this, where the object at the top has the old route and the destination.

```js
import { NextResponse,  } from 'next/server'

const redirectData = {
  "/old": {
    "destination": "/new",
    "permanent": true
  },
}

export async function middleware(request) {
  const pathname = request.nextUrl.pathname
  const redirectEntry = redirectData[pathname]

 
  if (redirectEntry ) {
    const statusCode = redirectEntry.permanent ? 308 : 307
    return NextResponse.redirect(request.nextUrl.origin + redirectEntry.destination, statusCode)
  }
 
  // No redirect found, continue without redirecting
  return NextResponse.next()
}
```

Not sure if @tonyboylehub has a different opinion how to do it.

L0STE: This worked as a wander! If @tonyboylehub doesn't have anything against it I think we should use this

MarkSackerberg (approved):

PR #178, opened 2024-08-07
removed unused page with little or zero text content

tonyboylehub:


src/components/products/candyMachine/index.js
```diff
@@ -199,7 +199,7 @@ export const candyMachine = {
     },
     /*
     {
-      ...recipesSection('candy-machine'),
```
MarkSackerberg, line 202: Since it is commented out... Why not remove it completely?

danenbm (approved):

MarkSackerberg (approved):

PR #180, opened 2024-08-08
Core Candy Machine - Fix all titles and descriptions

tonyboylehub:


src/pages/core-candy-machine/custom-guards/generating-client.md
```diff
@@ -1,7 +1,7 @@
 ---
-title: Generating Custom Guard Client
-metaTitle: Candy Machine - Generating Custom Guard Client
-description: How to generate a Umi compatible client for your custom guards.
+title: Generating Custom Guard Client for the new Core Candy Machine
+metaTitle: Core Candy Machine - Generating Custom Guard Client for newest Candy Machine, Core Candy Machine.
```
MarkSackerberg, line 3: this reads a bit complicated tbh. It's technically correct but it's probably easier for the users to read like this
```suggestion
metaTitle: Core Candy Machine - Generating Custom Guard Client for Core Candy Machine.
```

MarkSackerberg (dismissed):

PR #191, opened 2024-08-09
How to use the Local Validator -- General Guide

L0STE:


tonyboylehub:
We have Amman and should be promoting the use of that.

MarkSackerberg:
@L0STE iirc we agreed to merge this?

tonyboylehub:
You start with Windows installation then proceeded to abandon Windows by the time we get to `Creating a "Metaplex" Local Validator`.

Windows has no concept of `sudo`

src/components/products/guides/index.js
```diff
@@ -48,6 +48,18 @@ export const guides = {
...
+              href: '/guides/using-local-validator',
+              created: '2021-10-08',
+              updated: null, // null means it's never been updated
+            },
+            // {
+            //   title: 'Where do I start?',
+            //   href: '/guides/where-do-i-start-on-solana',
+            //   created: '2021-10-01',
+            //   updated: null, // null means it's never been updated
+            // },
```
MarkSackerberg, line 62: maybe remove the commented stuff

src/pages/guides/using-local-validator.md
```diff
@@ -0,0 +1,191 @@
...
+
+{% /dialect %}
+
+{% dialect title="Windows" id="Windows" %}
+
+```
+cmd /c "curl https://release.solana.com/v1.18.18/solana-install-init-x86_64-pc-windows-msvc.exe --output C:\solana-install-tmp\solana-install-init.exe --create-dirs"
+```
+
+{% /dialect %}
```
MarkSackerberg, line 34: this might be quite frustrating since later we say that it only works for windows when using WSL. Maybe we should recommend WSL here already insted of the .exe?

tonyboylehub: I made the same comment in a private conversation.

tonyboylehub:
There is the potential of adding the bash commands for windows with gitBash but would need to exspand on this in the doc.

src/pages/guides/using-local-validator.md
L0STE: So for this actually you can do everything up to when we create our own bash script to create your own validator with just the CLI (like dumping and loading programs) that's why I still left it because you can still actually run your local validator, will just not be that easy as calling `metaplex-local-validator`

src/components/products/guides/index.js
```diff
@@ -49,17 +49,11 @@ export const guides = {
               updated: null, // null means it's never been updated
             },
             {
-              title: 'Using the Local Validator',
-              href: '/guides/using-local-validator',
+              title: 'Setup a Local Validator',
+              href: '/guides/setup-a-local-validator',
               created: '2021-10-08',
```
MarkSackerberg, line 54: Last nitpick: you might want to update the date

MarkSackerberg (approved):

PR #196, opened 2024-08-13
Reworked and Merged `Getting Started` and `Connecting to Umi` docs

L0STE:
- Merged `Getting Started` and `Connecting to Umi` into `Getting Started`
- Improved readability and searchability of `Umi Basics`
- Added missing information


src/pages/umi/getting-started/index.md
```diff
@@ -53,10 +56,172 @@ export async function myFunction(
...
+
+{% totem-accordion title="Localnet" %}
+
+```ts
+const umi = createUmi('http://127.0.0.1:8899')
+```
+
+{% /totem-accordion %}
+
+{% /totem %}
```
tonyboylehub, line 132:
```suggestion

```

I would avoid this completely because more often than not these do not function well.
Would be better to advise people why not to use these and go checkout our RPC page instead and grab a free/paid RPC's from there.

src/pages/umi/getting-started/index.md
```diff
@@ -53,10 +56,172 @@ export async function myFunction(
...
+
+### Connecting a Wallet
```
tonyboylehub, line 134: The drop down accordian here has 3 titles

```
a new wallet
an existing wallet
a wallet adapter
```

this should be clearer. Existing wallets and walletadpater is essentially the same thing, both are existing wallets.

should be made clear these are file sytem wallets in some way to avoid potentially confusion.

tonyboylehub (commented):
We should also not this is not a complete walletAdapter solution and only includes the code to connection walletAdapter to umi if you have ALREADY installed and setup walletAdapter correctly which we should link out externally to currently and then switch that to an internal setup of walletAdapter one we've written that.

WalletAdapter setup guide external
https://github.com/anza-xyz/wallet-adapter/blob/master/APP.md

src/components/products/umi/index.js
```diff
@@ -26,8 +26,9 @@ export const umi = {
           title: 'Introduction',
           links: [
             { title: 'Overview', href: '/umi' },
-            { title: 'Getting started', href: '/umi/getting-started' },
-            { title: 'Connecting to Umi', href: '/umi/connecting-to-umi' },
```
MarkSackerberg, line 30: it would make sense to not just delete but redirect.

src/pages/umi/getting-started/index.md
```diff
@@ -53,10 +56,172 @@ export async function myFunction(
...
+// Tell Umi to use the new signer.
+umi.use(signerIdentity(signer))
+```
+
+{% /totem-accordion %}
+
+{% totem-accordion title="With an Existing Wallet" %}
+
+```ts
+import { generateSigner, createSignerFromKeypair } from '@metaplex-foundation/umi'
```
MarkSackerberg, line 159: The imports are not complete. createUmi, signerIdentity, fs is missing.

src/pages/umi/getting-started/index.md
```diff
@@ -4,40 +4,43 @@ metaTitle: Getting Started | Umi
...
 ```
 
-Then you want to tell Umi which wallet to use. This can either be a [keypair](/umi/connecting-to-umi#connecting-w-a-secret-key) or the [solana wallet adapter](/umi/connecting-to-umi#connecting-w-wallet-adapter).
+```
+npm i @metaplex-foundation/umi-bundle-defaults 
+```
 
-That's it, now pass your Umi instance around and add more plugins as needed.
+```
+npm i @metaplex-foundation/@solana/web3.js
```
MarkSackerberg, line 28: I don't think this is correct.

MarkSackerberg (changes requested):

MarkSackerberg (approved):

PR #203, opened 2024-08-19
Tony/sugar cli old guide

tonyboylehub:


tonyboylehub:
@MarkSackerberg do you happen to know where those guards examples for the config.json file are? I can't seem to find them.

src/components/products/candyMachine/index.js
danenbm, line 1: Can we also add a link to this guide at the main "Metaplex Program Guides" area that currently only has a Core section?  I think that's the easiest way to find guides now.

tonyboylehub: Yeah we can do that, I didn't add it there because of the potential clash with the othe "same guide" but we can get that bridge when it happens

tonyboylehub: Infact do we know what category we want to put it in? Shouldn't come under rust, or javascipt. General maybe?

tonyboylehub: I've pushed up  in a `general` catergory

danenbm: resolved offline, put in both Metaplex Program Guides > Candy Machine and direct top level under General for easy discoverability

danenbm (approved):

PR #206, opened 2024-08-21
Added a guide for Core Collection both using Anchor and Typescript

L0STE:
- Added `how-to-create-a-core-collection" as anchor guide (Following the template of PR 200, will adjust this too if we have some points on that PR)
- Added `how-to-create-a-core-collection" as javascript guide (Following the template made by Tony, some adjustment made so we can settle on that and change one of them if we want uniformation)
- Decided not to add it in the guide index but only in the left scroll (as it was done for the createAsset guide on typescript) to not overcrowd the section.

### DIFF
--- a/src/components/products/core/index.js
+++ b/src/components/products/core/index.js
@@ -23,47 +23,91 @@ export const core = {
       navigation: [
         {
           title: 'New!',
-          links: [{ title: 'JS SDK V1.0', href: '/core/core-js-sdk-v1-0' }],
+          links: [
+            { title: 'JS SDK V1.0', 
+              href: '/core/core-js-sdk-v1-0' }
+            ],
         },
         {
           title: 'Introduction',
           links: [
-            { title: 'Overview', href: '/core' },
+            { 
+              title: 'Overview', 
+              href: '/core' 
+            },
             {
               title: 'Getting Started',
               href: '/core/getting-started',
             },
-            { title: 'What is an Asset?', href: '/core/what-is-an-asset' },
+            { 
+              title: 'What is an Asset?', 
+              href: '/core/what-is-an-asset' 
+            },
             {
               title: 'Token Metadata Differences',
               href: '/core/tm-differences',
             },
-            { title: 'Ecosystem Support', href: '/core/ecosystem-support' },
-            { title: 'Anchor', href: '/core/using-core-in-anchor' },
-            { title: 'FAQ', href: '/core/faq' },
+            { 
+              title: 'Ecosystem Support', 
+              href: '/core/ecosystem-support' },
+            { 
+              title: 'Anchor', 
+              href: '/core/using-core-in-anchor' 
+            },
+            { 
+              title: 'FAQ', 
+              href: '/core/faq' 
+            },
           ],
         },
         {
           title: 'Features',
           links: [
-            { title: 'Creating Assets', href: '/core/create-asset' },
-            { title: 'Fetching Assets', href: '/core/fetch' },
-            { title: 'Updating Assets', href: '/core/update' },
-            { title: 'Transferring Assets', href: '/core/transfer' },
-            { title: 'Burning Assets', href: '/core/burn' },
+            { 
+              title: 'Creating Assets', 
+              href: '/core/create-asset' 
+            },
+            { 
+              title: 'Fetching Assets', 
+              href: '/core/fetch'
+            },
+            { 
+              title: 'Updating Assets', 
+              href: '/core/update' 
+            },
+            { 
+              title: 'Transferring Assets', 
+              href: '/core/transfer' 
+            },
+            { 
+              title: 'Burning Assets', 
+              href: '/core/burn' 
+            },
             {
               title: 'Collection Management',
               href: '/core/collections',
             },
-            { title: 'Helpers', href: '/core/helpers' },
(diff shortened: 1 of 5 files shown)


src/pages/core/guides/anchor/how-to-create-a-core-collection.md
```diff
@@ -0,0 +1,440 @@
...
+  .update_authority(update_authority.as_ref())
+  .system_program(&ctx.accounts.system_program.to_account_info())
+  .name(args.name)
+  .uri(args.uri)
+  .plugins(plugins)
+  .external_plugin_adapters(external_plugin_adapters)    
+  .invoke()?;
+```
+
+## Creating the Metadata for the Digital Asset
```
MarkSackerberg, line 247: Not sure if i would include it here - feels a bit confusing to go from rust to js with not much explanation why? Maybe use a dummy url and just link to one of the upload examples?

src/pages/core/guides/anchor/how-to-create-a-core-collection.md
```diff
@@ -0,0 +1,440 @@
+---
+title: How to Create a Core Collection
+metaTitle: How to Create a Core Collection | Core Guides
```
MarkSackerberg, line 3: SEO wise we should not use the same titles twice, so you might want to add "using Anchor" here and "using JavaScript" in the other file.

src/pages/core/guides/javascript/how-to-create-a-core-collection.md
```diff
@@ -0,0 +1,414 @@
...
+
+// run the wrapper function
+createCollection()
+```
+
+## Setting up Umi
+
+While setting up Umi you can use or generate keypairs/wallets from different sources. You create a new wallet for testing, import an existing wallet from the filesystem, or use `walletAdapter` if you are creating a website/dApp.  
+
+For this example we're going to set up Umi with a `generatedSigner()`.
```
MarkSackerberg, line 93: I'd just link the other options then

src/pages/core/guides/javascript/how-to-create-a-core-collection.md
```diff
@@ -0,0 +1,414 @@
...
+```js
+// Call upon umi's `uploadJson()` function to upload our metadata to Arweave via Irys.
+const metadataUri = await umi.uploader.uploadJson(metadata).catch((err) => {
+  throw new Error(err)
+})
+```
+
+Now we should finally have the URI of JSON file stored in the `metadataUri` providing it did not throw any errors.
+
+### Minting the Core Collection
```
MarkSackerberg, line 274: in the anchor example you show how to use plugins. Maybe we should show it in JS, too?

MarkSackerberg (changes requested):
Hi @L0STE Some nitpicks.
The title needs to be changed for sure, otherwise they will appear as issues in semrush

src/pages/core/guides/anchor/how-to-create-a-core-collection.md
tonyboylehub: Agreed, this would be in a program enviroment so upload should have already happened. I feel like the upload part of this is uneeded.

If it was a local Rust script example I could see the point of uploading the image via Rust, but uneeded to for Anchor. Just needs to passed in as a param arg.

src/components/products/core/index.js
```diff
@@ -179,6 +179,10 @@ export const core = {
         {
           title: 'Javascript',
           links: [
+            {
+              title: 'How to Create a Core Collection',
+              href: '/core/guides/javascript/how-to-create-a-core-collection',
+            },
```
tonyboylehub, line 185: SEO

src/components/products/core/index.js
```diff
@@ -192,6 +196,10 @@ export const core = {
         {
           title: 'Anchor',
           links: [
+            {
+              title: 'How to Create a Core Collection',
+              href: '/core/guides/anchor/how-to-create-a-core-collection',
+            },
```
tonyboylehub, line 202: SEO

tonyboylehub:
Debating on whether the Anchor stuff should mention CPI, it probably should.

src/pages/core/guides/anchor/how-to-create-a-core-collection-with-anchor.md
```diff
@@ -0,0 +1,349 @@
...
+```rust
+let update_authority = match &ctx.accounts.update_authority {
+    Some(update_authority) => Some(update_authority.to_account_info()),
+    None => None,
+};
+```
+
+After preparing all the necessary accounts, we pass them to the `CreateCollectionV2CpiBuilder` and use `.invoke()` to execute the instruction, or `.invoke_signed()` if we need to use signer seeds.
+
+For more details on how the Metaplex CPI Builder works, you can refer to this [documentation](/guides/rust/how-to-cpi-into-a-metaplex-program#using-metaplex-rust-transaction-cpi-builders)
```
L0STE, line 201: @tonyboylehub I added it here, do you think we should make it more visible/explain how CPI works or this is fine?

tonyboylehub: I like that. Thanks!

MarkSackerberg (approved):

PR #213, opened 2024-09-04
fix: use of owner instead of mint

jim4067:


MarkSackerberg:
You are totally right. Thank you very much!

MarkSackerberg (approved):

PR #224, opened 2024-09-15
Core CM UI guide

MarkSackerberg:
Overview on which functions are needed and which approach to take when building your own core candy machine.

Not a start to end guide finishing with a working website but pointers in the right direction. I don't feel we should start teaching React here.

Two discussion points from my side:
1. Do we want to have a very similar guide in the TM Candy Machine section? Then we would have a guide how to create the CM and how to build the UI. We do not have "How to create the CM" for core. Also it might be seen as duplicate content SEO wise since besides the package name it's technically the same
2. Do we want to add links to existing community made UIs (for core currently just mine as far as i know) 

I've also done some changes to other pages where we missed information:
- `signAllTransactions` in umi
- How to fetch guard related account to get the status


src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+{% /totem-accordion  %}
+
+{% /dialect %}
+{% /dialect-switcher %}
+
+### Fetch additional Candy Machine related Accounts
+The choice of Guards you implement may necessitate fetching additional accounts. For instance, if you plan to verify a wallet's minting eligibility and are utilizing the `mintLimit` Guard, you would need to retrieve the `mintCounter` account. This account maintains a record of how many NFTs a particular wallet has already minted under that specific guard.
+
+#### `MintLimit` Accounts
+When the [`MintLimit`](/core-candy-machine/guards/mint-limit) guard is active, it's advisable to retrieve the `MintCounter` for the user's wallet. This allows you to check whether the user has reached their minting limit or if they're still eligible to mint more.
```
tonyboylehub, line 431:
```suggestion
When the [`MintLimit`](/core-candy-machine/guards/mint-limit) guard is active, it's advisable to retrieve the `MintCounter` account for the user's wallet. This allows you to check whether the user has reached their minting limit or if they're still eligible to mint additional items.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+const mintCounter = await safeFetchMintCounterFromSeeds(umi, {
+  id: 1, // The mintLimit id you set in your guard config
+  user: umi.identity.publicKey,
+  candyMachine: candyMachine.publicKey,
+  candyGuard: candyMachine.mintAuthority,
+});
+```
+
+#### `NftMintLimit` Accounts
+Similar to the `MintLimit` guard it can make sense to fetch the `NftMintCounter` of the [`NftMintLimit`](/core-candy-machine/guards/nft-mint-limit) guard to be able to verify eligibility.
```
tonyboylehub, line 447:
```suggestion
Similar to the `MintLimit` guard it can make sense to fetch the `NftMintCounter` account of the [`NftMintLimit`](/core-candy-machine/guards/nft-mint-limit) guard to verify eligibility.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+The following code snippet demonstrates how to fetch the `NftMintCounter`. Note that this example assumes you've already obtained the Candy Machine and Candy Guard data:
```
tonyboylehub, line 449:
```suggestion
The following code snippet demonstrates how to fetch the `NftMintCounter` account. Note that this example assumes you've already obtained the Candy Machine and Candy Guard data:
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+  mint: asset.publicKey, // The address of the nft your user owns
+  candyGuard: candyMachine.mintAuthority,
+  candyMachine: candyMachine.publicKey,
+});
+      
+const nftMintCounter = fetchNftMintCounter(umi, pda)
+```
+
+#### `AssetMintLimit` Accounts
+Similar to the `NftMintCounter` guard it can make sense to fetch the `AssetMintCounter` of the [`AssetMintLimit`](/core-candy-machine/guards/asset-mint-limit) guard to be able to verify eligibility.
```
tonyboylehub, line 468:
```suggestion
Similar to the `NftMintCounter` guard it can make sense to fetch the `AssetMintCounter` account of the [`AssetMintLimit`](/core-candy-machine/guards/asset-mint-limit) guard to verify eligibility.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+The following code snippet demonstrates how to fetch the `AssetMintCounter`. Note that this example assumes you've already obtained the Candy Machine data:
```
tonyboylehub, line 470:
```suggestion
The following code snippet demonstrates how to fetch the `AssetMintCounter` account. Note that this example assumes you've already obtained the Candy Machine data:
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+  asset: asset.publicKey, // The address of the core nft your user owns
+  candyGuard: candyMachine.mintAuthority,
+  candyMachine: candyMachine.publicKey,
+});
+
+const assetMintCounter = fetchAssetMintCounter(umi, pda);
+```
+
+#### `Allocation` Accounts
+For the `Allocation` guard it can make sense to fetch the `AllocationTracker` to verify that more NFTs can be minted from a given group.
```
tonyboylehub, line 489:
```suggestion
For the `Allocation` guard it can make sense to fetch the `AllocationTracker` account to verify that additional NFTs can be minted from a given group.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+The following code snippet demonstrates how to fetch the `AllocationTracker`. Note that this example assumes you've already obtained the Candy Machine data:
```
tonyboylehub, line 491:
```suggestion
The following code snippet demonstrates how to fetch the `AllocationTracker` account. Note that this example assumes you've already obtained the Candy Machine data:
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+const allowListProof = await safeFetchAllowListProofFromSeeds(umi, {
+  candyGuard: candyMachine.mintAuthority,
+  candyMachine: candyMachine.publicKey,
+  merkleRoot: getMerkleRoot(allowlist),
+  user: umi.identity.publicKey,
+});
+```
+
+### Fetch Wallet Data
+To validate the legibility you may also want to fetch information about the connected wallet. Depending on the Guards you are using you may want to know how much SOL is in the wallet and which Tokens and NFTs they own.
```
tonyboylehub, line 533:
```suggestion
To validate the legibility you may also want to fetch information about the connected wallet. Depending on the Guards you are using you may want to know how much SOL is in the wallet and which Tokens and NFTs the wallet owns.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+To fetch the SOL balance the built in `getAccount` umi function can be used to fetch the wallet account:
+```ts
+const account = await umi.rpc.getAccount(umi.identity.publicKey);
+const solBalance = account.lamports;
```
tonyboylehub, line 538: don't know how you want to do this but there's actually a `getBalance` function which might make more sense. Open on this.

```ts
const walletBalance = await umi.rpc.getBalance(umi.identity.publicKey)
const solBalance = walletBalance.basisPoints;
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+const umi = createUmi('<ENDPOINT>').use(dasApi());
+
+const assets = await umi.rpc.getAssetsByOwner({
+    umi.identity.publicKey
+});
+
+```
+
+## Verify legibility
+After fetching all this data it can be used to check if a connected wallet is allowed to mint.
```
tonyboylehub, line 558:
```suggestion
After fetching all the required data you can then verify if the connected wallet is allowed to mint or not.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+It's important to note that when groups are attached to a Candy Machine, the `default` guards apply universally across all groups. However, in this scenario, you must utilize the specific groups for minting, as minting directly from the `default` group is no longer an option.
```
tonyboylehub, line 560:
```suggestion
It's important to note that when groups are attached to a Candy Machine, the `default` guards apply universally across all created groups.  Also, when groups are enabled the ability to mint from the `default` group becomes disabled and you must use the created groups for minting.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+Therefore, if there are no groups defined you need to check if all the mint conditions of the `default` group are met. If there are groups defined the combination of `default` guards and each group need to be validated.
```
tonyboylehub, line 562:
```suggestion
Therefore, if there are no groups defined you need to check if all the mint conditions of the `default` group are met. If there are groups defined the combination of both the `default` guards and the current minting group guards both need to be validated.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+Given a Candy Machine with the `startDate`, `SolPayment` and `mintLimit` guards attached that is not leveraging groups the following validations should be done before allowing the user to call the mint function. It is assumed that the `candyGuard` was fetched before and one Core NFT Asset should be minted.
+
+1. Validate the `startDate` is in the past. Note that we are not using the users devices time here but instead fetch the current internal Solana Blocktime since this is the time the Candy Machine will use for the validation on mint:
```
tonyboylehub, line 566:
```suggestion
1. Validate the `startDate` is in the past. Note that we are not using the users device time here but instead fetching the current internal Solana Blocktime since this is the time the Candy Machine will use for the validation on mint: 
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+    // mintCounter PDA exists (not the first mint)
+    if (mintCounter && mintLimit.limit >= mintCounter.count
+    ) {
+      allowed = false;
+    }
+}
+```
+
+When the wallet is not eligible to mint it is helpful to disable the mint button and show the User the reason for it. E.g. a `Not enough SOL!` message.
```
tonyboylehub, line 624:
```suggestion
When a wallet is not eligible to mint it is helpful to disable the mint button and show the user the reason for not being eligible to mint. E.g. a `Not enough SOL!` message.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+## Guard Routes
+Certain Guards require specific instructions to be executed before minting can occur. These instructions create accounts that store data or provide proof of a wallet's eligibility to mint. The execution frequency of these instructions varies depending on the Guard type. 
+
+{% callout type="note" title="Target Audience of this section" %}
+In case you are not using the `Allocation`, `FreezeSolPayment`, `FreezeTokenPayment` or `Allowlist` guard it is safe to skip this section.
+{% callout type="note" title="Configuration" %}
+
+Some Guards need their routes executed only once for the entire Candy Machine. For these, it's not necessary to include a function in the UI but can be run upfront once through a script:
```
tonyboylehub, line 633:
```suggestion
Some Guards need their routes executed only once for the entire Candy Machine. For these, it's not necessary to include a functional interaction (button) in the UI but can be run on page load once through a script:
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+- [Allocation](/core-candy-machine/guards/allocation)
+- [FreezeSolPayment](/core-candy-machine/guards/freeze-sol-payment)
+- [FreezeTokenPayment](/core-candy-machine/guards/freeze-token-payment)
+
+Other Guards require their routes to be executed for each individual wallet. In these cases, the route instruction should be run prior to the mint transaction:
+- [Allowlist](/core-candy-machine/guards/allow-list)
+
+For an example of how to implement Guard routes, consider the case of the Allowlist guard. This assumes that the allowListProof has been fetched as described earlier, and that `allowlist` represents an array of eligible wallet addresses. The following code demonstrates how to handle this scenario in your implementation.
```
tonyboylehub, line 641:
```suggestion
For an example of how to implement Guard routes, consider the case of the **Allowlist** guard. This assumes that the `allowListProof` has been fetched as described earlier, and that `allowlist` represents an array of eligible wallet addresses. The following code demonstrates how to handle this scenario in your implementation.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+      path: "proof",
+      merkleRoot: getMerkleRoot(allowlist),
+      merkleProof: getMerkleProof(allowlist, publicKey(umi.identity)),
+    },
+  })
+}
+```
+
+## Create a Mint Function
+It is recommended to implement legibility checks for all the guards that are attached. Keep in mind that if there are any groups attached the `default` ones apply to all groups, but you also have to use the groups and can not mint from the `default` group anymore.
```
tonyboylehub, line 670:
```suggestion
It is recommended to implement legibility checks for all the guards that are attached. Keep in mind that if there are any groups attached the `default` guards will apply to all additional groups, while simultaneously disabling the `default` group.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+After those checks are done and, if required, the route instructions were run the mint transaction can be built. Depending on the guards `mintArgs` have to be passed in. These are arguments that help building the mint transaction by passing in the correct Accounts. For example the `mintLimit` accounts needs the `mintCounter` account. The Umi SDK abstracts these details away but still requires some information to build the transaction correctly.
```
tonyboylehub, line 672:
```suggestion
After those checks are done and, if required, the route instructions were run the mint transaction can be built. Depending on the guards, `mintArgs` may have to be passed in. These are arguments that help build the mint transaction by passing in the correct accounts and data. For example the `mintLimit` guard requires the `mintCounter` account. The Umi SDK abstracts these details away but still requires some information to build the transaction correctly.
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
```diff
@@ -0,0 +1,789 @@
...
+
+Depending on the specific setup it can be helpful to allow minting multiple NFTs in one Transaction by combining the [Transaction Builders](/umi/transactions#transaction-builders).
+
+```ts
+let builder = transactionBuilder()
+  .add(mintV1(...))
+  .add(mintV1(...))
+```
+
+In case you allow to mint many Assets in one transaction you might experience `Transaction too large` errors. The function [`builder.fitsInOneTransaction(umi)`](/umi/transactions#transaction-builders) allows to check for this before sending the transaction so that the transaction can be split before sending it. In case splitting is needed using [`signAllTransactions`](/umi/transactions#building-and-signing-transactions) is recommended so that only one popup has to be approved in the Wallet Adapter.
```
tonyboylehub, line 735:
```suggestion
If you add to many `mintV1` instructions into a transaction you will recieve a `Transaction too large` error. The function [`builder.fitsInOneTransaction(umi)`](/umi/transactions#transaction-builders) allows to check for this before sending the transaction so that the transaction can be split before being sentt. In case splitting is needed using [`signAllTransactions`](/umi/transactions#building-and-signing-transactions) is recommended so that only one popup has to be approved in the Wallet Adapter.    
```

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
MarkSackerberg: This is not correct from my POV. You only run it once independent of the UI.

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
MarkSackerberg: Mhm, this does not really make a difference, does it? `getBalance` might load less data maybe, but for a normal wallet there should not be much of a difference

MarkSackerberg:
Thank you @tonyboylehub! If you could have a look at the last two comments that would be awesome.

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
tonyboylehub: yep ignore me then :D

src/pages/core-candy-machine/guides/create-a-core-candy-machine-ui.md
tonyboylehub: just terminology. I'm not fussed either way, both works. One function just explains it more explicitly. If it's a deprecated function also no point in changing it.

MarkSackerberg: Resolving so that i do not have to retest :)

tonyboylehub (approved):

PR #230, opened 2024-09-24
update: use generated function

jim4067:


MarkSackerberg (approved):
Agree, this change makes sense. Thank you!

PR #240, opened 2024-10-21
Added pNFT examples to Token Metadata

tonyboylehub:
Reworked a few areas of bunch of pages to add more details surrounding pNFTs and the additional accounts they need for transacting with Token Metadata program.


src/components/products/tokenMetadata/index.js
```diff
@@ -64,7 +64,7 @@ export const tokenMetadata = {
               href: '/token-metadata/delegates',
             }, // Delegate + Revoke + Delegated transfer and burn.
             { title: 'Locking Assets', href: '/token-metadata/lock' },
-            { title: 'Programmable NFTs', href: '/token-metadata/pnfts' },
+            { title: 'Programmable NFTs (pNFT)', href: '/token-metadata/pnfts' },
```
danenbm, line 67: Nit: I think technically you should do pNFTs since the word you were turning into an acronym was plural?

src/pages/token-metadata/burn.md
```diff
@@ -42,7 +44,7 @@ await burnV1(umi, {
 
 If the asset that you are trying to burn is part of a collection you additionally need to pass the collectionMetadata account into the function:
 
-{% dialect-switcher title="Burning Assets that are part of a collection" %}
+{% dialect-switcher title="NFT Asset Burn that's part of a collection" %}
```
danenbm, line 47: Why this change?  The grammar I think is actually more correct in text being replaced?

wdyt about "Burning NFT Assets that are part of a collection"?

src/pages/token-metadata/burn.md
```diff
@@ -59,3 +61,53 @@ await burnV1(umi, {
...
+import {
+  burnV1,
+  fetchDigitalAssetWithAssociatedToken,
+  findMetadataPda,
+  TokenStandard,
+} from "@metaplex-foundation/mpl-token-metadata";
+import { publicKey, unwrapOption } from "@metaplex-foundation/umi";
+import { base58 } from "@metaplex-foundation/umi/serializers";
+
+// The NFT mint ID
```
danenbm, line 80: Should this say "pNFT"?

src/pages/token-metadata/burn.md
```diff
@@ -59,3 +61,53 @@ await burnV1(umi, {
...
+const mintId = publicKey("11111111111111111111111111111111");
+
+// Fetch the NFT Asset with the Token Account
```
danenbm, line 83: Should this say "pNFT Asset"?

src/pages/token-metadata/burn.md
```diff
@@ -59,3 +61,53 @@ await burnV1(umi, {
...
+const assetWithToken = await fetchDigitalAssetWithAssociatedToken(
+  umi,
+  mintId,
+  umi.identity.publicKey
+);
+
+// Determine if the Asset is in a collection
```
danenbm, line 90: Also should this say "pNFT Asset"?

src/pages/token-metadata/lock.md
```diff
@@ -60,3 +103,45 @@ await unlockV1(umi, {
...
+```ts
+import {
+    fetchDigitalAssetWithAssociatedToken,
+    TokenStandard,
+    unlockV1
+} from "@metaplex-foundation/mpl-token-metadata";
+import { publicKey } from "@metaplex-foundation/umi";
+import { base58 } from "@metaplex-foundation/umi/serializers";
+
+// Mint ID of the Asset
```
danenbm, line 120: Should this say "pNFT Asset"?

src/pages/token-metadata/pnfts.md
```diff
@@ -1,27 +1,27 @@
 ---
-title: Programmable NFTs
-metaTitle: Programmable NFTs | Token Metadata
+title: Programmable NFTs (pNFTS)
+metaTitle: Programmable NFTs (pNFTS) | Token Metadata
```
danenbm, line 3: Nit: Lowercase "s" so like "pNFTs" is more proper for this case

src/pages/token-metadata/pnfts.md
```diff
@@ -1,27 +1,27 @@
...
 
-A good example of why this can be problematic is that Token Metadata cannot enforce secondary sales royalties. Even though the royalty percentage is stored on the **Metadata** account, it is up to the user or program that performs the transfer to decide whether they want to honour it or not. We talk more about this and how PNFTs fix this issue [in a section below](#use-case-royalty-enforcement).
+A good example of why this can be problematic is that Token Metadata cannot enforce secondary sales royalties. Even though the royalty percentage is stored on the **Metadata** account, it is up to the user or program that performs the transfer to decide whether they want to honour it or not. We talk more about this and how pNFTs fix this issue [in a section below](#use-case-royalty-enforcement).
 
-Programmable NFTs are introduced to solve this issue in a flexible way that **allows creators to customise the authorization layer** of their assets.
+Programmable NFTs are introduced to solve this issue in a flexible way that **allows creators to customize the authorization layer** of their assets.
```
danenbm, line 15: Interesting detail choice we've never discussed afaik.   But since have chosen to change "customize" to USA English.  So then should "honour" be changed to  "honor"?

src/pages/token-metadata/pnfts.md
```diff
@@ -30,9 +30,9 @@ Let's dive into these two abilities in more detail.
 
 ## More granular delegates
 
-Since all PNFTs operations must go through the Token Metadata program, it can create a new delegate system on top of the spl-token delegate. One that is more granular and allows PNFT owners to pick the operations they want to delegate to a third party.
+Since all pNFTs operations must go through the Token Metadata program, it can create a new delegate system on top of the spl-token delegate. One that is more granular and allows pNFT owners to pick the operations they want to delegate to a third party.
 
-Information for this new delegate system is stored on a special **Token Record** PDA that is derived from both the Mint and the Token accounts of the PNFT. When a new delegate authority is assigned to a PNFT, the Token Metadata program synchronizes that information on both the Token account and the Token Record account.
+Information for this new delegate system is stored on a special **Token Record** PDA that is derived from both the Mint and the Token accounts of the pNFT. When a new delegate authority is assigned to a pNFT, the Token Metadata program synchronizes that information on both the Token account and the Token Record account.
```
danenbm, line 35: Thank you for fixing all these!  Great cleanup, getting the docs nicer and nicer each iteration.

src/pages/token-metadata/update.md
```diff
@@ -57,3 +125,56 @@ await updateV1(umi, {
...
+import {
+  collectionToggle,
+  fetchMetadataFromSeeds,
+  TokenStandard,
+  updateAsUpdateAuthorityV2,
+} from '@metaplex-foundation/mpl-token-metadata'
+import { publicKey, unwrapOptionRecursively } from '@metaplex-foundation/umi'
+
+// The Mint ID of the pNFT Asset
+const mintId = publicKey('HckzL4iRK3j6DsznzedwNFzmFj4z7xCBoRkMMrd8THd5')
```
danenbm, line 148: Should this just be like '11111111' like your other examples rather than a real Pubkey?

src/pages/token-metadata/update.md
```diff
@@ -6,17 +6,85 @@ description: Learn how to update Assets on Token Metadata
...
-- **Primary Sale Happened**: A boolean that indicates whether the asset has been sold before.
-- **Is Mutable**: A boolean that indicates whether the asset can be updated again. When changing this to `false`, any future updates will fail.
-- **Collection**: This attribute enables us to set or clear the collection of the asset. Note that when setting a new collection, the `verified` boolean must be set to `false` and [verified using another instruction](/token-metadata/collections).
-- **Collection Details**: This attribute enables us to set or clear the collection details of the asset.
-- **Rule Set**: This attribute enables us to set or clear the rule set of the asset. This is only relevant for [Programmable Non-Fungibles](/token-metadata/pnfts).
+## Updatable Fields
 
 Note that certain delegated authorities can also update the **Metadata** account of assets as discussed in the "[Delegated Authorities](/token-metadata/delegates)" page.
```
danenbm, line 12: I really like the way you broke each attribute/parameter out and explained it in a mini-section.  Easier to read in the rendered version.

One slight thing I could think of would be to add some kind of text in the "##Updatable Fields" section indicating that these are the individual attributes that can be provided to the updateV1 method.

There is already a statement lower between the two code examples, and it does become obvious once you see the code examples below, but if you are looking at this the first time, you just see the data object, primarySaleHappened, isMutable, etc. and you don't get the "establishing shot" until you see the updateV1 code example.  Might be slightly confusing for a newcomer.  

I think just saying something like "Here's the different attributes that can be sent to the updateV1 method" or something like that would work.

danenbm (commented):
Great cleanup, left a couple minor comments

src/pages/token-metadata/pnfts.md
tonyboylehub: Yeah it was just one I spotted on the fly. I should probably go through with a find and replace on english to american english as I don't think we've ever done a deep dive on that. I know I'm one to get them mixed up occasionally.

PR #242, opened 2024-10-21
added truncate as default to all getAssetWithProof() examples

tonyboylehub:
```ts
const assetWithProof = await getAssetWithProof(umi, assetId, 
    { truncateCanopy: true }
);
```
to all examples.


danenbm (approved):
LGTM

PR #280, opened 2024-12-10
Feat/api components

tonyboylehub:
**Preview**
https://developer-hub-git-feat-api-components-metaplex-foundation.vercel.app/aura

Online Sandboxs for code testing
- Csharp https://dotnetfiddle.net/
- Go https://www.boot.dev/playground/go
- Java https://www.jdoodle.com/online-java-compiler
- PHP https://www.tehplayground.com/
- Pyhton 3.x.x + tested locally
- Ruby couldn't find online play ground with http enabled.
- Swift couldn't find online play ground with http enabled.
- Rust tested locally.


src/components/apiComponents/languageSelector.jsx
```diff
@@ -0,0 +1,46 @@
...
+  { name: 'python', icon: 'PythonIcon' },
+  { name: 'ruby', icon: 'RubyIcon' },
+  { name: 'go', icon: 'GoIcon' },
+  { name: 'java', icon: 'JavaIcon' },
+  { name: 'csharp', icon: 'CSharpIcon' },
+  { name: 'swift', icon: 'SwiftIcon' },
+  { name: 'kotlin', icon: 'KotlinIcon' },
+  { name: 'php', icon: 'PhpIcon' },
+  // { name: 'typescript', icon: 'TypescriptIcon' },
+  { name: 'shell', icon: 'ShellIcon' },
```
danenbm, line 14: How is 'shell' different than 'curl'?  I don't see 'shell' doing anything on the rendering right now.

src/components/apiComponents/endPointSelector.jsx
```diff
@@ -0,0 +1,94 @@
...
+    name: 'Eclipse Mainnet',
+    uri: 'https://aura-eclipse-mainnet.metaplex.com',
+    value: 'eclipseAuraMainnet',
+  },
+  custom: {
+    name: 'Custom',
+    uri: 'custom',
+    value: 'custom',
+  },
+}
```
danenbm, line 17: Assuming we are fine with showing the URLs, right now when you select 'Custom', you get the box showing 'https://aura-mainnet.metaplex.com' that you can modify.  But when you select 'Eclipse Mainnet', it just shows the selector box (as this is a fixed URL).  It would be nicer if it showed the Eclipse Mainnet URL but greyed out.  So that you could see what its actually using.

src/pages/aura/api/v1/das/get-asset.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Get Asset
+metaTitle: Get Asset Method | Aura API
+description: Learn about the Get Asset Aura API Method.
+---
+
+Fetch a NFT Digital Asset from the Solana blockchain including NFTs, pNFTs, cNFTs.
```
danenbm, line 7: Nit: I would say "Fetch _an_ NFT Digital Asset" wdyt?

src/components/apiComponents/exampleSelector.jsx
```diff
@@ -0,0 +1,57 @@
...
+  selectedExample,
+  handleSetExample,
+}) => {
+  return (
+    <div className="w-full">
+      <label
+        className="mb-3 text-sm font-medium text-gray-800 dark:text-neutral-400"
+        htmlFor="endPoint"
+      >
+        Loaded Example
```
danenbm, line 16: Should this be called like "Preloaded Example"?  Not sure about exact terminology.

src/components/apiComponents/exampleSelector.jsx
```diff
@@ -0,0 +1,57 @@
...
+      </label>
+      <div className="relative flex w-full gap-2">
+        <div className="relative flex h-12 w-full">
+          <Select
+            onChange={(e) => handleSetExample(e.target.value)}
```
danenbm, line 21: When a Loaded Example is set, should it set the endpoint to Solana mainnet (since that's where the examples are), and in the future if you add Eclipse loaded examples it could select Eclipse mainnet?

src/components/apiComponents/endPointSelector.jsx
```diff
@@ -0,0 +1,94 @@
...
+  eclipseAuraMainnet: {
+    name: 'Eclipse Mainnet',
+    uri: 'https://aura-eclipse-mainnet.metaplex.com',
+    value: 'eclipseAuraMainnet',
+  },
+  custom: {
+    name: 'Custom',
+    uri: 'custom',
+    value: 'custom',
+  },
```
danenbm, line 16: Is it a problem for us to show 'https://aura-mainnet.metaplex.com' right now when they select "Custom"?  For example we only show the beta signup on the Solana RPC page.

src/components/apiComponents/endPointSelector.jsx
tonyboylehub: I like that!

src/components/apiComponents/exampleSelector.jsx
tonyboylehub: I didn't use the term 'pre' because I load nothing on page load at the moment. So the Loaded Example is nothing, nothing is load.

Do we want to pre load an example into active state?

src/components/apiComponents/exampleSelector.jsx
tonyboylehub: I hadn't decided on this yet because we don't have aura mainnet ready for launch possibly with these examples. So we'll need to decide on how we are loading these. Either potentially with a custom with basic solana api endpoint. Or we let me people fill in their own and navigate.

This is why I ended up with the assets being in a group under "Solana Mainnet" to point this out.

src/components/apiComponents/endPointSelector.jsx
tonyboylehub: I don't think it's ready for beta/testing? The end points were removed from the main docs so I assumed they wouldn't go live with this either.

src/components/apiComponents/languageSelector.jsx
tonyboylehub: Yeah I can ditch shell.

src/components/apiComponents/languageSelector.jsx
```diff
@@ -0,0 +1,46 @@
+import { Icon } from '@/components/icons'
+
+const languages = [
```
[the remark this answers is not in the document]
tonyboylehub: Yeah I'm happy to sort these as needed. I'll chat with you for order in slack.

src/components/apiComponents/exampleSelector.jsx
tonyboylehub: I haven't adjusted this at the moment incase people are using custom endpoints. Might be pretty annoying if everytime you select and example it swaps out your custom. Though I guess if "custom" then don't change maybe.

danenbm: i think it would be good to pick Solana mainnet if nothing is picked, but not change it if they have something picked, if that is easy to do.
Contingent on whether we are publishing the mainnet endpoint yet

tonyboylehub: Added current behaviour:

If custom endpoint is selected and an example is selected DO NOT change the enpoint value from custom
else change endpoint value to match example.

danenbm (approved):
This is awesome great job LGTM!

PR #283, opened 2024-12-17
fix/navigation update and UI fixes

tonyboylehub:
https://developer-hub-git-tony-navigation-update-metaplex-foundation.vercel.app/

**Changes**

- Navigation Update for MPL and Aura
     - Moved all programs into MPL Category
- Removed ugly secondary mobile menu that no one really knew how to access
- Updated main mobile menu to provide two extra links
     - Home
     - Programs and Tools (this replaces the ugly menu in point 2)
- Added the extra social icons
- Remove the big unusable search input box that didn't do anything other than be a button and replaced with icon.
- Added the custom scrollbar we were using in all overflow components to the main body as well to match.
- Fixed column rendering from 3 -> 2 depending on screensize to 2 -> 1 which gives a much nicer layout for mobile.


MarkSackerberg:
Looking good to me on mobile!
Just one question for the desktop Version: Should we add a "Home" link in the desktop version, too? Like it is now users need to know the metaplex logo is "home". (It's for most websites but not always)

MarkSackerberg (dismissed):

danenbm (approved):

PR #290, opened 2024-12-22
docs: Update mint without Guards and mint with Guards example

coderhyy:


MarkSackerberg:
Thanks for catching that the example is the wrong way around!

MarkSackerberg (approved):

PR #291, opened 2024-12-23
hybrid guide - extend scripts with boilerplate

MarkSackerberg:
I feel like more boilerplate is better here.
Especially because of the required `publicKey as publicKeySerializer` in the import.


tonyboylehub:
Thanks for this Mark. Can we also add `MPL-404 Hybrid` program to the list of Program Guides in the Guides navigation as in this PR to save opening a second on?

MarkSackerberg:
Done!

tonyboylehub (approved):

PR #307, opened 2025-01-13
Token Claimer Guide

L0STE:
Added the Token Claimer Guide with Merkle Tree but without concurrency (Explained the risks of this approach with locking accounts)

Link: https://developer-hub-git-token-claimer-smar-29dce6-metaplex-foundation.vercel.app/token-metadata/guides/anchor/token-claimer-smart-contract

### DIFF
--- a/src/pages/token-metadata/guides/anchor/token-claimer-smart-contract.md
+++ b/src/pages/token-metadata/guides/anchor/token-claimer-smart-contract.md
@@ -35,6 +35,55 @@ Parent2 = Hash(Hash(C) + Hash(D))
 Root = Hash(Parent1 + Parent2)
 ```
 
+{% diagram %}
+
+{% node #root label="Root Node" theme="slate" /%}
+{% node #root-hash label="Hash" parent="root" x="56" y="40" theme="transparent" /%}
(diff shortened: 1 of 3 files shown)


src/pages/token-metadata/guides/anchor/token-claimer-smart-contract.md
```diff
@@ -0,0 +1,972 @@
...
+pub struct UserReceipt {
+  pub user: Pubkey,
+}
+```
+
+{% /totem-accordion %}
+
+{% /totem %}
+
+**Note**: The `user_receipt` account could be left empty to minimize rent costs. However, this makes it harder to check on the frontend if a user has already claimed their tokens. This tradeoff should be considered based on the specific requirements of the program.
```
MarkSackerberg, line 521: why does it make it harder? Based on the seeds one could check if the account or not, so no data in the account would be required?

MarkSackerberg (commented):
Thank you. This is looking great!
Just some minor nits and we are good to go.

MarkSackerberg (approved):
One last minor nit that you might want to look into: The text and the title do not mention the word "airdrop" even though thats the term people might be familiar with. (And also technically not correct)
You might want to add it somewhere

PR #313, opened 2025-01-21
migrate getting started page to sdk page

tonyboylehub:


MarkSackerberg (changes requested):
- The old "getting started" page still exists but is not pointed to anymore (file not deleted but middleware added)
- why is there no rust page anymore / why is it not reachable anymore like in the old getting started page?
- we should update the "getting started" entry in the menu on the left and not just route through the middleware

MarkSackerberg (approved):

PR #335, opened 2025-03-31
Add note about RedeemAmount parameter being a index

BHEEEET:


MarkSackerberg:
Hey,
thank you for the suggestion. Personally I feel like this makes it harder to understand though.

PR #340, opened 2025-04-02
update guard list

MarkSackerberg:
Adding missing guards to the available guards list
https://developer-hub-git-updatecorecmguardsection-metaplex-foundation.vercel.app/core-candy-machine/guards#all-available-guards


tonyboylehub (changes requested):
Thanks Mark.

We got one 404 on the `sol-fixed-price` link. But other than that looks great. :)

tonyboylehub (approved):
Thanks!

PR #353, opened 2025-04-14
fixed CreateV1CpiBuilder invoke_signed example

danielpavel:
CreateV1CpiBuilder was being called with invoke while the example was mentioning invoke_signed.


tonyboylehub:
hi @danielpavel.

I'm going to close this out as I fixed this in another larger PR. Thanks for the input though and sorry for the slow reply.

danielpavel:
hi @tonyboylehub, glad it's fixed. Keep up the good work!

PR #357, opened 2025-04-19
guides spring cleaning

tonyboylehub:


src/pages/guides/validators.md
```diff
@@ -3,11 +3,94 @@ title: Validators and Staking
...
+- Software: Solana validator software and supporting tools
+- SOL Tokens: To cover vote transaction fees and optional self-stake
+
+### Validator Economics
+
+Validators earn rewards through:
+
+1. **Transaction Fees**: A portion of transaction fees paid by users
+2. **Inflation Rewards**: New SOL tokens distributed to validators and delegators
+3. **MEV (Maximal Extractable Value)**: Additional value that can be extracted by influencing transaction ordering
```
MarkSackerberg, line 44: Not sure if we should add stuff like that. While it might be the case now it also might change in the future which would require us to update it

src/pages/guides/what-is-solana.md
```diff
@@ -34,6 +34,36 @@ The Solana blockchain is a high-performance, decentralized blockchain platform d
 7. **Low Costs:**
    Solana offers low transaction fees, making it an attractive option for developers and users looking to build and interact with dApps and DeFi platforms without the high costs associated with other blockchains.
 
-   ## Upcoming
-   
-   In addition to what is already a performant chain Solana is advancing with new developments like Runtime v2, which aims to improve transaction processing and a simplified developer experience for everyone.
+## Upcoming Developments
+
+In addition to what is already a performant chain, Solana is advancing with new developments like Runtime v2, which aims to improve transaction processing and provide a simplified developer experience for everyone.
+
```
MarkSackerberg, line 40: Same here. If we frame it as upcoming we need to remember to update it later.

MarkSackerberg (approved):

MarkSackerberg (approved):

PR #363, opened 2025-05-15
Playground api

tonyboylehub:
Removal of API selector in the DAS Playground
https://developer-hub-git-playground-api-metaplex-foundation.vercel.app/das-api/methods/get-asset

Defaults to `https://api.devnet.solana.com` but saves your supplied RPC address to local storage and recalled upon subsequant page visits.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,302 +1,198 @@
-lockfileVersion: '6.0'
+lockfileVersion: '9.0'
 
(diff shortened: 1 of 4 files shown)


danenbm (commented):
LGTM but could we have one devnet example so they can see something working on devnet?

danenbm (approved):
Approved with request to add devnet example in follow-on PR

PR #367, opened 2025-06-03
added devnet examples to dropdown

tonyboylehub:
- create and added devnet examples to dropdown
- fixed some navigation issues from moving pages over.

https://developer-hub-git-das-playground-dev-baf570-metaplex-foundation.vercel.app/das-api/methods/get-asset

### DIFF
--- a/src/components/apiComponents/exampleSelector.jsx
+++ b/src/components/apiComponents/exampleSelector.jsx
@@ -1,12 +1,15 @@
-import { Select } from '@headlessui/react'
-import { ChevronDownIcon } from '@heroicons/react/20/solid'
-import clsx from 'clsx'
+import { Select } from '@headlessui/react';
+import { ChevronDownIcon } from '@heroicons/react/20/solid';
+import clsx from 'clsx';
 
 const ApiExampleSelector = ({
   examples,
   selectedExample,
   handleSetExample,
 }) => {
+  const mainnetExamples = examples.filter((example) => example.chain === 'solanaMainnet');
+  const devnetExamples = examples.filter((example) => example.chain === 'solanaDevnet');
+
   return (
     <div className="w-full">
       <label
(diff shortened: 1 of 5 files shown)


src/components/products/das-api/index.js
```diff
@@ -39,10 +36,10 @@ export const das = {
             { title: 'Get Asset Proof', href: '/das-api/methods/get-asset-proof' },
             { title: 'Get Asset Proofs', href: '/das-api/methods/get-asset-proofs' },
             { title: 'Get Asset Signatures', href: '/das-api/methods/get-asset-signatures' },
-            { title: 'Get Asset By Authority', href: '/das-api/methods/get-asset-by-authority' },
-            { title: 'Get Asset By Creator', href: '/das-api/methods/get-asset-by-creator' },
-            { title: 'Get Asset By Group', href: '/das-api/methods/get-asset-by-group' },
-            { title: 'Get Asset By Owner', href: '/das-api/methods/get-asset-by-owner' },
+            { title: 'Get Assets By Authority', href: '/das-api/methods/get-assets-by-authority' },
```
MarkSackerberg, line 39: While these title and url changes are technically correct we should add the old ones to the middleware to not break something

src/components/apiComponents/exampleSelector.jsx
```diff
@@ -18,24 +21,37 @@ const ApiExampleSelector = ({
...
               '*:text-black dark:text-white'
             )}
           >
-            <option value={-1}>-</option>
-            <optgroup label="Solana Mainnet">
-              {examples.map((example, index) => {
-                return (
-                  <option key={index} value={index}>
+            <option value="">-</option>
+            {mainnetExamples.length > 0 && (
```
MarkSackerberg, line 37: Since devnet is the default url I feel like the devnet examples should be at the top and mainnet at the bottom?
Might make it less confusing since you can just pick the first from the list?

danenbm: +1

danenbm (commented):
LGTM except agree devnet examples could be at top of list since its the default endpoint

PR #375, opened 2025-07-14
Program Guide Index

tonyboylehub:
Added a guide index with selectable tags for basic filtering.

Preview: https://developer-hub-git-guides-guides-index-metaplex-foundation.vercel.app/guides


MarkSackerberg (approved):
Approved.
A auto generated index would have been even nicer, but probably doesn't make sense to invest the time for it.

PR #387, opened 2025-08-19
fix asset collection update

tonyboylehub:
Updating the asset collection was coded as the old way. Updated.

### DIFF
--- a/src/pages/core/update.md
+++ b/src/pages/core/update.md
@@ -124,6 +124,8 @@ import {
   update,
   fetchAsset,
   fetchCollection,
+  collectionAddress,
+  updateAuthority
 } from "@metaplex-foundation/mpl-core";
 
 const assetId = publicKey("11111111111111111111111111111111");
@@ -143,6 +145,7 @@ const updateTx = await update(umi, {
   asset,
   collection,
   newCollection: publicKey("22222222222222222222222222222222"),
+  newUpdateAuthority: updateAuthority('Collection', [newCollectionId]),
 }).sendAndConfirm(umi);
 ```
 
--- a/src/pages/core/update.md
+++ b/src/pages/core/update.md
@@ -124,22 +124,26 @@ import {
   update,
   fetchAsset,
   fetchCollection,
-  updateAuthority,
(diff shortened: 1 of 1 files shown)


src/pages/core/update.md
```diff
@@ -124,22 +124,26 @@ import {
...
 } from "@metaplex-foundation/mpl-core";
 
 const assetId = publicKey("11111111111111111111111111111111");
+
 const asset = await fetchAsset(umi, assetId);
-const oldCollectionId = publicKey("22222222222222222222222222222222");
-const collection = await fetchCollection(umi, oldCollectionId);
-const newCollectionId = publicKey("33333333333333333333333333333333");
+
+const collectionId = collectionAddress(asset)
```
MarkSackerberg, line 133: The import of `collectionAddress` is missing

src/pages/core/update.md
```diff
@@ -124,22 +124,26 @@ import {
   update,
   fetchAsset,
   fetchCollection,
-  updateAuthority,
 } from "@metaplex-foundation/mpl-core";
 
 const assetId = publicKey("11111111111111111111111111111111");
+
 const asset = await fetchAsset(umi, assetId);
```
MarkSackerberg, line 131: not a change you did but a top level await is invalid. I've started to wrap the snippets with the following to not have an error thrown directly. Same for the missing umi initialization.

```
(async () => {
    ....
})();
```

it introduces additional boilerplate and might be resolved with Keith' different way to import snippets, so no hard feelings on if you have to change it now or later.

MarkSackerberg (changes requested):
Thank you!
There is one minor change required at least (adding the missing import)

src/pages/core/update.md
tonyboylehub: I'm actually started to think maybe we split the code out into imports and then snippets in seperate blocks. None of these code examples run through straight copy and paste anyway because of needing the UMI setup. I'm not a fan of wrapping the functions/code to appease top level await. Makes things untidy. I don't think I've ever seen that in other docs.

PR #388, opened 2025-08-19
Update resize language

blockiosaurus:


MarkSackerberg (approved):

michalzajda:
@MarkSackerberg can you provide backward compatibility? How to create metadata with the old size 679?

This optimization broke attestation flow in Wormhole bridge. No new token with metadata can be bridged with Wormhole as of now (in Solana Mainnet only)

FYI, https://github.com/wormhole-foundation/wormhole/issues/4478

PR #417, opened 2025-12-29
Add redirects + fix internal linking

tonyboylehub:
- Add redirects for old localized path structure
- Fixed internal links for localization new paths

redirects: `ja/umi -> ja/dev-tools/umi` etc...


MarkSackerberg (approved):
Did not test all the links but the ones I tested are working fine.

PR #426, opened 2026-01-06
Genesis Presale

MarkSackerberg:
This adds docs for the Genesis Presale function including translations.

Pending Confirmation on the fees page.

### DIFF
--- a/src/components/products/genesis/index.js
+++ b/src/components/products/genesis/index.js
@@ -33,14 +33,14 @@ export const genesis = {
     },
     presale: {
       deposit: {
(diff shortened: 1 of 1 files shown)


tonyboylehub:
My previous one and this are very different and we should align on the structure.

For example 
- mine already assumes you have setup the initial Gensesis account from the getting started page.
- my version displays the fee's yours tells the user to visit the fees page for fees.
- yours reads more like an e2e guide where mine is split into the sections for that particular step.

I'm not fussed how we do it we just need to align on it.

PR #444, opened 2026-02-10
Initial Solana Basics Guides

MarkSackerberg:
Adding more "Solana Basics" guide.
They can be accessed by clicking "Guides" in the top menu. Added back in for now, but happy for other proposals.

New Pages are marked as "new" here: https://developer-hub-git-guides-solana-basics-metaplex-foundation.vercel.app/guides


tonyboylehub:
This looks great. I think the index page needs redoing but that's somrthing I can look into if you want.

tonyboylehub (commented):
THis looks really good and thank you for the metadata generation. 

Couple of small comments and can we change the folder name from /guides to /solana?

We should also add a 'Create a Solana Token' page and document both Solana CLI and why the Metaplex CLI/our SDK is is better (because it handles automatic metadata creation etc...). Then link it out to the /tokens 'create token' page.

PR #455, opened 2026-02-23
Add Skill to developer tools

MarkSackerberg:
Even though besides the installation information not much is required in here i think adding it to the developer hub could increase discoverability


tonyboylehub (changes requested):
3 pages have duplicate h1 and h2 headings.

PR #466, opened 2026-03-03
fix: Space calculation error in oracle-plugin-example

skartik-sk:
There was Small error while calculating space in oracle-plugin-example docs.

Given : 
8+5+1

Changed 
8+5+1 +1 

+1 for vault_bump:u8

### DIFF
--- a/src/pages/en/smart-contracts/core/guides/oracle-plugin-example.md
+++ b/src/pages/en/smart-contracts/core/guides/oracle-plugin-example.md
@@ -110,7 +110,7 @@ pub struct Oracle {
     pub vault_bump: u8,
 }
 impl Space for Oracle {
-    const INIT_SPACE: usize = 8 + 5 + 1;
+    const INIT_SPACE: usize = 8 + 5 + 1+1;
 }
 ```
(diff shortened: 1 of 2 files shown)


tonyboylehub:
Verified — you're right, and thanks for catching it.

`Oracle` has two `u8` bump fields but `INIT_SPACE` only counted one:

```
8  Anchor discriminator
5  OracleValidation
1  bump
1  vault_bump   <- missing
--
15 (the constant said 14)
```

This isn't just an arithmetic nit either: `CreateOracle` passes `space = Oracle::INIT_SPACE` to `init`, so anyone following the guide creates the account a byte short and serialization fails. The guide's own prose confirms both bumps are intentional — *"We save the bump for both the Oracle PDA and the Oracle Vault PDA"* — so the struct was right and the constant was wrong.

We've adopted a policy of not merging branches from forks directly, so I've reimplemented the fix on an internal branch and credited you in the commit and PR description:

#559

Closing this in favour of that one. The `en` and `ja` pages are both covered, same as yours. The find is entirely yours — thank you for reporting it.

PR #482, opened 2026-03-24
fix: update developer-hub paths to include /docs

AtoMicKraK1n:
### Description
Fixed broken navigation in the Solana section where links were missing the /docs prefix, causing incorrect redirects to the homepage.

### Changes:

- Updated internal paths to include /docs.

- Verified links now resolve correctly instead of redirecting to metaplex.com.

### DIFF
--- a/public/llms.txt
+++ b/public/llms.txt
@@ -62,14 +62,22 @@
 
 ## Genesis (Token Launch)
 
-- [Aggregation API](https://metaplex.com/docs/smart-contracts/genesis/aggregation): Public API for querying Genesis launch data by genesis address or token mint. Includes on-chain state fetching.
-- [API](https://metaplex.com/docs/smart-contracts/genesis/api): Public API for querying Genesis launch data by genesis address or token mint. No authentication required.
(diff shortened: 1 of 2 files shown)


tonyboylehub:
Thanks for flagging this — you were right that the guide index links were broken. The component renders a plain `<a href>`, which Next does not prefix with the configured `basePath` (`/docs`), so all 62 links were 404ing.

Two things came out of digging into it:

- Hardcoding `/docs` into each path produces `/ja/docs/...` for non-English locales, because `getLocalizedPath` prepends the locale afterwards — so the fix worked for English but broke the other three locales. Using `next/link` lets Next apply `basePath` itself and keeps the ordering correct.
- While in there we also found the guide data was keyed `jp`/`kr` while `LocaleContext` yields `ja`/`ko`, so Japanese and Korean guide names were silently falling back to English.

The `public/llms.txt` changes in this PR are also auto-generated at build time by `scripts/generate-llms-txt.mjs`, which is why this branch kept picking up conflicts there.

We've since adopted a policy of not merging fork branches directly. The fix has landed on an internal branch covering both bugs:

#557

Closing this in favour of that one — but thank you for spotting the original problem, it had been live in all four locales for a while.

AtoMicKraK1n:
Glad it helped!

PR #518, opened 2026-04-22
Add locked LP tokens documentation page for Genesis

tonyboylehub:
New concept page explaining how LP tokens are permanently program-locked during bonding curve graduation. Includes verification code using the Genesis SDK to check the RaydiumCpmmBucketV2 account and confirm the lpLockSchedule startCondition is set to Never.


src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -0,0 +1,214 @@
...
+  - Raydium CPMM liquidity
+proficiencyLevel: Intermediate
+programmingLanguage:
+  - JavaScript
+  - TypeScript
+faqs:
+  - q: Are LP tokens burned or locked during graduation?
+    a: LP tokens are permanently program-locked, not burned. They are transferred to an associated token account owned by a Genesis bucket signer PDA. The bucket's lpLockSchedule has its startCondition set to Never, meaning no wallet can ever claim them.
+  - q: Can anyone withdraw the locked LP tokens?
+    a: No. The lpLockSchedule on the RaydiumCpmmBucketV2 account has its startCondition set to Never. There is no instruction, authority, or upgrade path that can release them. They are permanently program-locked.
```
danenbm: This is perhaps a bit too strong to say.  There is no instruction or authority that can release them.  But its probably a bit too much to say there's no upgrade path, because the program is not immutable.  And permanent might even be too strong here.  I think the idea is that we would have to have something like a governance vote before we changed what we did with them.  I think it could be sufficient to just say "There is no instruction or authority that can release them".

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -0,0 +1,214 @@
...
+---
+
+LP tokens created during [bonding curve graduation](/smart-contracts/genesis/bonding-curve#lifecycle) are permanently program-locked in a Genesis-owned bucket. No wallet can ever withdraw them. {% .lead %}
+
+## Summary
+
+When a Genesis bonding curve sells out and graduates into a [Raydium CPMM](/smart-contracts/genesis/bonding-curve-theory#phase-3-graduated) pool, the resulting LP tokens are transferred to a PDA owned by the Genesis program. The vesting schedule on this bucket is set to `Never`, making the tokens permanently inaccessible.
+
+- **LP tokens are not burned** — they are transferred to a Genesis bucket signer PDA with vesting set to never
+- **Permanently program-locked** — the `lpLockSchedule.startCondition` is `Never`, so no instruction can release them
```
danenbm: Maybe just say "program locked" here instead of permanently, as the program is not immutable.

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -0,0 +1,214 @@
...
+*Maintained by Metaplex · Last verified April 2026 · [Genesis](https://github.com/metaplex-foundation/mpl-genesis)*
+
+## How LP Token Locking Works
+
+During graduation, the Genesis program creates a Raydium CPMM pool using the accumulated SOL and a portion of the token supply. Raydium returns LP tokens representing the liquidity position. These LP tokens are transferred to an associated token account (ATA) owned by a **bucket signer PDA** — a program-derived address that the Genesis program controls but has no private key.
+
+The `RaydiumCpmmBucketV2` account stores the lock configuration in its `extensions.lpLockSchedule` field. This is a `ClaimSchedule` whose `startCondition` is set to `{ __kind: 'Never' }`. Because the claim start condition can never be satisfied, no instruction can withdraw the LP tokens from the bucket signer's ATA.
+
+{% callout type="note" %}
+Some platforms refer to this as "burning" LP tokens. In Genesis, the LP tokens are not sent to a burn address — they remain in a verifiable onchain account. The term **permanently program-locked** is more accurate because the tokens exist and can be audited, but no wallet can ever access them.
```
danenbm: Same here would soften the language.

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -0,0 +1,214 @@
...
+When LP tokens are permanently program-locked, the output confirms:
+
+```
+LP lock start condition: Never
+LP token balance: 123456789
+Bucket signer (LP token owner): <PDA address>
+Bucket signer ATA (holds LP tokens): <ATA address>
+```
+
+The `startCondition.__kind` value of `Never` is the definitive proof that the LP tokens cannot be withdrawn. This condition is set at graduation and cannot be modified after the fact.
```
danenbm: I think you also need to read cliff condition, and that has to also be Never.  so that there's no start to vesting and also no cliff.

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -0,0 +1,214 @@
...
+| `cliffAmountBps` | `number` | Cliff unlock percentage in basis points (irrelevant when start is `Never`) |
+
+{% callout type="note" %}
+The `duration`, `period`, `cliffCondition`, and `cliffAmountBps` fields are present in the `ClaimSchedule` struct but are functionally irrelevant when `startCondition` is `Never`. The lock cannot begin, so no vesting schedule ever executes.
+{% /callout %}
+
+## Notes
+
+- LP tokens are locked automatically during graduation — no separate instruction is required
+- The lock is permanent and irreversible — there is no admin key, upgrade path, or governance mechanism that can change the `startCondition` from `Never` to any other value
```
danenbm: We would use governance if we ever wanted to change it but I think its best to leave out the claims on permananence.  I would check everywhere and not make such strong claims of its permanence, just that its program locked (see other comment)

danenbm (changes requested):
Overall looks good and the example is great.  Made a few comments.  Also it does feel like it repeats the content a couple times, seems like it could be more concise.

Lastly, seems like we still need somewhere calling out that with launchpool there is in fact a vesting schedule that is quarterly.  Maybe you would want that on another page under launchpool?  But seems like there could be one page covering both if you made this one more concise on the bonding curve part.  At the same time, having a dedicated bonding curve one might be best for the intent of this page, so maybe you are right to isolate the bonding curve content here.

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -25,36 +25,34 @@ programmingLanguage:
...
-    a: Fetch the RaydiumCpmmBucketV2 account using the Genesis SDK and check that extensions.lpLockSchedule has startCondition.__kind set to Never. The lpTokenBalance field shows the exact number of LP tokens held.
+    a: Fetch the RaydiumCpmmBucketV2 account using the Genesis SDK and check that extensions.lpLockSchedule has both startCondition.__kind and cliffCondition.__kind set to Never. The lpTokenBalance field shows the exact number of LP tokens held.
   - q: What is the difference between burning LP tokens and program-locking them?
-    a: Burning sends tokens to an inaccessible address and removes them from circulation. Program-locking transfers them to a PDA with vesting set to never — the tokens still exist onchain and can be verified, but no wallet can ever withdraw them. The end result is the same — the liquidity is permanent.
+    a: Burning sends tokens to an inaccessible address and removes them from circulation. Program-locking transfers them to a PDA with vesting set to never — the tokens still exist onchain and can be verified, but no wallet can withdraw them. The end result is the same — the liquidity is locked.
```
danenbm, line 34: Nit: burning could be an SPL token burn

src/pages/en/smart-contracts/genesis/locked-lp-tokens.md
```diff
@@ -157,58 +163,51 @@ Key fields on the `RaydiumCpmmBucketV2` account relevant to LP token locking:
...
+Fetch the `RaydiumCpmmBucketV2` account using the Genesis SDK and check that `extensions.lpLockSchedule` has both `startCondition.__kind` and `cliffCondition.__kind` set to `Never`. The `lpTokenBalance` field shows the exact number of LP tokens held. See [Verifying the LP Token Lock](#verifying-the-lp-token-lock) for the full code example.
 
 ### What is the difference between burning LP tokens and program-locking them?
 
-Burning sends tokens to an inaccessible address and removes them from circulation. Program-locking transfers them to a PDA with vesting set to `Never` — the tokens still exist onchain and can be verified, but no wallet can ever withdraw them. The end result is the same: the liquidity is permanent.
+Burning sends tokens to an inaccessible address and removes them from circulation. Program-locking transfers them to a PDA with vesting set to `Never` — the tokens still exist onchain and can be verified, but no wallet can withdraw them. The end result is the same: the liquidity is locked.
```
danenbm, line 201: same comment as above

danenbm (approved):
Approved LGTM, with one comment

PR #544, opened 2026-07-29
Feat/deprecated programs nav

tonyboylehub:


MarkSackerberg (approved):
image
Approved with one nit: it might make sense to remove now duplicated depreciation warnings like we have with gumdrop

PR #554, opened 2026-09-04
docs(solana): interactive PLGRND flows for PDAs, priority fees and transaction anatomy

gymnasy55:
## What

Proposal: give three Solana Basics pages a small interactive example right where the concept is explained, so readers can try it instead of only reading about it. The examples are flows from [PLGRND](https://plgrnd.io), an open-source visual playground for Solana (MIT, source: https://github.com/RedDuckTeam/PLGRND), embedded with the existing `{% video … embed=true /%}` Markdoc tag. No new components, no new dependencies.

- **Understanding PDAs** — derives the Token Metadata PDA for a mint (`"metadata"` + program id + mint), the exact computation behind `findMetadataPda`. Paste any mint and the metadata address and bump update.
- **Compute Units and Priority Fees** — a live fee calculator under "Example Calculation": CU limit × price ÷ 1,000,000 plus the 5,000-lamport base fee.
- **Solana Transaction Fundamentals** — a real mainnet transaction fetched with `getTransaction`, shown next to the "Key Components" table: signatures, header, account keys, recent blockhash, instructions. Paste any signature to inspect it.

## Also fixed

"Example Calculation" on the priority fees page converted 200,000,000 micro-lamports to 200,000 lamports; the correct value is 200 lamports (1 lamport = 1,000,000 micro-lamports). The example now uses 10,000 micro-lamports per CU → 2,000 lamports = 0.000002 SOL, and adds the 0 lamports), matching the "Fee Calculation"section above it.                                                                                                    
## Details

- Same tag and URL in `en`, `ja`, `ko`, `zh`. The fee and transaction pages are English in all four locales already, so only the tag was added there; the new PDA `ja`/`ko`/`zh` (a draft translation, happy to take corrections).
- `updated` bumped in the frontmatter and in `src/components/products/guides/index.js`.

Two things I'd defer to you on: `Video.jsx` gives the iframe no `title` or `sandbox`, and adds a `Powered by PLGRND` badge that is part of the embed. If you'd prefer a dedicated `{% plgrnd %}` tag with `sandbox`, `title` and `loading="lazy"`, or would rather drop any of the three flows, I'll adjust.

### DIFF
--- a/src/components/products/guides/index.js
+++ b/src/components/products/guides/index.js
@@ -39,7 +39,7 @@ export const guides = {
               title: 'Solana Transaction Fundamentals',
               href: '/solana/solana-transaction-fundamentals',
               created: '02-04-2026',
-              updated: null,
+              updated: '09-03-2026',
             },
             {
               title: 'Solana Programs',
@@ -51,7 +51,7 @@ export const guides = {
               title: 'Understanding PDAs',
               href: '/solana/understanding-pdas',
               created: '2021-10-01',
-              updated: null,
+              updated: '09-03-2026',
             },
             {
               title: 'Validators and Staking',
@@ -81,7 +81,7 @@ export const guides = {
               title: 'Compute Units and Priority Fees',
               href: '/solana/compute-units-and-priority-fees',
               created: '02-04-2026',
-              updated: null,
+              updated: '09-03-2026',
             },
           ],
         },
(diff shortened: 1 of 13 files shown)


blockiosaurus:
Hello and thank you for the PR! While we appreciate the contribution, our internal policy prohibits linking external embeds like this directly from our docs in case the underlying content changes or is compromised. But I'd be happy to discuss utilizing some of the content as directly hosted assets.

PR #555, opened 2026-09-09
docs(agents): replace Nori prototype docs with Metaplex x402

tonyboylehub:
## Summary

The `/agents/nori` pages documented an early service-agent prototype (`agent-plumber`) that was never released publicly and is no longer maintained. This replaces them with documentation for the service that actually shipped, **Metaplex x402**.

Reported as [metaplex-foundation/mpl-agent#21](https://github.com/metaplex-foundation/mpl-agent/issues/21) — the source link on those pages pointed at `metaplex-foundation/agent-plumber`, which 404s.

The broken link turned out to be the smallest problem. The pages described:

| Nori docs said | Reality |
|---|---|
| Settled in **SOL** vs. agent PDA, Jupiter spot conversion | **USDC** |
| Base URL `NORI_URL` | Never published; service is `https://api.metaplex.com/x402` |
| `/v1/chat/completions`, `/v1/images/generations`, `/v1/solana/rpc` | `/x402/chat/completions`, `/x402/images/generations`, `/x402/rpc` |
| Bearer token from `POST /auth/handshake`, 15-minute TTL | x402 payment header; 24-hour JWT for delegated agents |
| Delegate-pay primary, x402 fallback | x402 primary, three payment modes |
| `/a2a` JSON-RPC + `/.well-known/agent-card.json` | Both 404 — never shipped |
| "The source repository is open source" | Server is not public; the **client** is, at `metaplex-foundation/x402` |

There is also a name collision worth noting: **`metaplex.com/nori` is now the Metaplex AI copilot**, an unrelated product. Keeping a `/docs/agents/nori` URL describing paid inference would have pointed two different products at the same word on the same domain.

## Changes

**Added** `src/pages/en/agents/x402/`:

- `index.md` — overview: services, Mech as the operating agent, quick start, payment-mode comparison, pricing model, FAQ, glossary
- `payment-modes.md` — guide: prerequisites, funding requirements, all three modes with code, delegation approve/revoke, reactive vs. proactive integration, common errors
- `api-reference.md` — reference: endpoints, discovery, per-service code, 402 challenge fields, full pricing tables, scheme options, client exports, env vars

**Removed** the four `/agents/nori` pages across `en`, `ja`, `ko`, `zh` (16 files).

**Rewired** the agents product nav and the top-level nav menu; regenerated `llms.txt` via `scripts/generate-llms-txt.mjs`.

**Redirects** — 16 entries in `src/middleware.js`, 4 paths × 4 locales:

| From | To |
|---|---|
| `/agents/nori` | `/agents/x402` |
| `/agents/nori/delegate-to-nori` | `/agents/x402/payment-modes` |
| `/agents/nori/pricing-and-billing` | `/agents/x402/api-reference` |
| `/agents/nori/example-agents` | `/agents/x402/payment-modes` |

**Corrected** `agent-commerce.md`, which stated "Metaplex doesn't ship an x402 server or client" — we now ship both. Its FAQ body, `faqs` frontmatter, and the "by Flag, Not Stub" heading were updated to match.

## Verification

Endpoints, pricing tables, and protocol constants were read from the live service on 2026-09-08 rather than transcribed from the README. The 402 challenge section is decoded from a real `PAYMENT-REQUIRED` header: protocol v2, `exact` scheme, mainnet CAIP-2 network ID, USDC mint, the `$0.001` minimum as `amount: "1000"`, the `metaplex:x402:chat.completions` memo, and the `X-METAPLEX-CORE-ASSET` delegation header.

- `pnpm build` passes; all three routes prerender
- All 16 redirects verified returning `308` against a built server
- `markdownlint` clean
- JSON-LD verified per page type — `FAQPage` on both guides, `HowTo` on payment-modes, `TechArticle`-only on the reference

## Locale parity

All three pages exist in `ja`, `ko`, and `zh`, mirroring the English structure line for line. Prose, headings, and `title`/`metaTitle`/`description`/`faqs`/`howToSteps` frontmatter are translated; `keywords`, `about`, dates, protocol constants, and code blocks stay in English per the localisation rules. `validate-translations` is back to baseline — the only remaining gaps are six pre-existing `test-*` pages.

Every H2/H3 in the locale pages carries an **explicit English heading id**. The site slugifies heading text with `@sindresorhus/slugify`, which strips CJK entirely:

| Heading | Slug |
|---|---|
| `標準的なSolanaウォレットで支払う` | `solana` |
| `委任されたエージェントで即時に支払う` | *(empty → falls back to `section`)* |
| `표준 Solana 지갑으로 결제하기` | `solana` |
| `使用标准 Solana 钱包支付` | `solana` |

Without explicit ids, translated headings collide on `solana` or fall back to `section`, breaking every in-page jump link. This also keeps anchors identical across locales, so cross-locale links resolve. The `{% #id %}` annotation is an existing convention here — 24 locale files already use it.

That work also surfaced a real bug in the English reference: the link to `#metaplexsvmexactscheme-options` never resolved, because slugify splits the camelCase heading into `metaplex-svm-exact-scheme-options`. Fixed with an explicit id.

The `agent-commerce.md` correction is propagated to `ja`, `ko`, and `zh` as well.

### Verification

- `pnpm build` passes; all **12** x402 routes prerender (3 pages × 4 locales)
- All 9 locale pages return `200` and all **16** nori redirects return `308` to a live page, checked against a built server
- **28** anchor links across 4 locales resolve, with no duplicate or empty heading ids
- `markdownlint` clean across all 16 touched pages
- JSON-LD verified per page type in every locale — `FAQPage` on both guides, `HowTo` on payment-modes, `TechArticle`-only on the reference


tonyboylehub:
Worked through all six review comments. Verified each against the pages before changing anything — all six were valid, and two turned out to be broader than reported.

**Locale links pointed at the English pages.** `MarkdocLink` hands the href straight to `next/link`, which prepends `basePath` but *not* the locale, so a root-relative `/agents/x402` from a `ja` page resolved to the English page. The review flagged 9 links in `agent-commerce`; auditing the rest of the files this PR touches found **129** across all four x402 and agent-commerce locale pages. All prefixed now — and I confirmed all 48 distinct targets exist in each locale first, so none of them turn an English fallback into a 404.

**Both Summary sections described only the sign-every-payment flow.** `api-reference` said the client *"signs a USDC payment and retries"*, which is wrong for a delegated agent — it authenticates once with a message signature and the service builds payments from the agent wallet. `payment-modes` said wiring *"differs only in which payment scheme you register"*, which skips the auth token store and the client extension or fetch wrapper delegated mode also needs. Both corrected.

**`agent-commerce` opened its x402 section by defining the generic protocol.** Since chunks are retrieved in isolation, that section could be extracted without conveying that Metaplex operates an x402 service at all. Now leads with the service, then the protocol.

**Added a warning about browser token persistence.** The page recommended `LocalStorageMetaplexCoreExecuteDelegateAuthTokenStore` for browser code with no caveat, but that JWT authorizes payments from the agent's wallet for 24 hours and any XSS on the origin can read and replay it. Now recommends the in-memory store by default, with minimal balances and prompt revocation.

**First `DAS` mention now linked** — flagged for `ja`/`ko`, but `en` and `zh` had it too. The first mention is in the endpoint table on line 55 while the canonical link only appeared on line 154. All four fixed.

### Verification

- `pnpm build` exit 0, all 12 x402 routes prerender
- `markdownlint` clean across the 16 touched files, callout tags balanced
- **Zero** English-target links left in the rendered locale HTML
- Sampled locale targets confirmed resolving to real built pages

### One note for reviewers

The `~35%` of locale links repo-wide that lack a locale prefix (roughly 2,000 of 5,700) is a pre-existing issue beyond these files — left alone here.

Separately, the explicit `{% #id %}` anchors this PR adds to locale headings are a workaround for slugify dropping CJK. #556 fixes that at the root, so once it lands these annotations become belt-and-braces rather than load-bearing. They're harmless to keep and still useful for keeping anchors identical across locales.

PR #556, opened 2026-09-14
fix(i18n): give CJK headings real anchor ids

tonyboylehub:
Closes DEV-167.

`@sindresorhus/slugify` transliterates to ASCII and drops whatever it cannot map, so **every** Japanese, Korean, and Chinese heading rendered with an empty `id`. That silently broke the "On this page" TOC, in-page jump links, and heading anchor copy-links across 91% of localized pages. English was unaffected and nothing failed the build, which is why it went unnoticed since localized docs were introduced.

## The fix

Adds `src/shared/slugifyHeading.js` as the single source of truth for heading ids, used by all three call sites (`usePage`, `Heading`, `HeadingAnchors`) so the TOC, the server-rendered id, and the client-side anchor finally agree.

Pure-ASCII headings keep the existing `slugify` behaviour so indexed English deep links stay byte-identical. Headings containing non-ASCII fall back to a Unicode-preserving GitHub-style slug, which is what authored in-page links already assume.

Two refinements beyond the original ticket analysis, both found while measuring:

- **Variation selectors survived `\p{M}`.** `⚙️` is `⚙` + U+FE0F, and keeping the selector after the emoji is stripped left a stray leading hyphen, shifting 4 English emoji headings. Now stripped explicitly.
- **Typographic punctuation forced the Unicode path.** An em dash or curly quote pushed otherwise-English headings off the legacy slugifier. Folding those to ASCII first took English drift from an estimated ~18 headings down to 2.

Dedupe is also restored; the counter previously did nothing for CJK because every id was the same empty string.

## Measured across `src/pages`

| | before | after |
|---|---|---|
| Empty heading ids | 8,618 | **0** |
| Duplicate ids | — | **0** |
| Broken in-page links | 1,114 | 453 |
| English slugs changed | — | **2 of 5,266** |

The 2 changed English slugs are both genuinely mixed-script headings in `solana/translated/japanese.md`, and nothing links to either.

## Also included

- Fixes 3 jump links in the Genesis `api-client` page that never resolved.
- Adds `scripts/audit-heading-anchors.mjs` (`pnpm lint:anchors`) and a GitHub Actions workflow, since the repo had no `.github/` at all.

## On the baseline file

The remaining 453 broken links are recorded in `scripts/heading-anchors-baseline.json` so CI blocks *new* breakage without failing on the existing backlog. 358 of them point at anchors that do exist on the matching `/en` page — localized pages whose headings were translated while their jump links still target the English anchors. That is a content fix across ~135 locale files and wants its own PR.

## Verification

- `pnpm lint` clean, full production build exit 0
- Rendered `<h2 id="よくある質問">` with TOC ids matching exactly
- English ids confirmed byte-identical
- Audit exits 1 on a deliberately broken anchor, 0 when clean


tonyboylehub:
Worked through all seven. Every one held up, and two were real bugs in what this branch added — thanks CodeRabbit.

## Repeated h4–h6 headings rendered duplicate ids

`parseTableOfContents` only counted `h2`/`h3`, so everything else fell through to the stateless slug in `Heading.jsx`. `dev-tools/cli/config/wallets.md` rendered **four** headings with `id="arguments"` and four with `id="example"`. 28 pages affected.

Worse, `parseTableOfContents` is skipped entirely when a page sets `tableOfContents: false` — so none of those **64** pages got counted ids at all.

And the audit had been reporting zero duplicates the whole time, because it applied a counter to every level while the renderer did not. It was measuring a model the site never implemented, which is exactly the kind of false assurance a CI gate shouldn't give.

Split id assignment out into `assignHeadingIds`, which runs for every page with content. It assigns in **two passes, h2/h3 first**, so the headings that reach the TOC claim their slug before `h1` and `h4`–`h6` and keep the ids they already had. The second pass shares the counter, so a repeated `#### Notes` dedupes against the `## Notes` it previously collided with — rather than the other way round, which a single pass would have done.

| | |
|---|---|
| h2/h3 slugs that shift | **0** of 5,266 |
| Built pages with duplicate or empty heading ids | **0** of 2,053 |

`wallets.md` now renders `arguments`, `arguments-1`, `arguments-2`, `arguments-3`; `explorer.md` keeps `#notes` on the `## Notes` h2 and gives the `#### Notes` h4 `notes-1`. The audit models the same two passes so the two can't drift apart again.

## A second occurrence of a baselined link passed CI

Confirmed exactly as described — the comparison counted only entries already pushed to `newBroken`, so with a baseline of 1, two occurrences both computed `seen = 1` and neither was flagged. Counts now live in a separate map; two occurrences of a new link report both, and the run exits 1.

## The audit aborted on input it should just report

Both latent (neither case exists in the repo today), both worth fixing:

- A `~~~` block containing a literal ``` closed the fence early, after which code was parsed as prose. Fences now close only on a matching delimiter of at least the opening length.
- `](#rate%off)` threw `URIError` out of `decodeURIComponent` and killed the run before it could report anything. Now caught and treated as a broken anchor.

## Workflow hardening

Added `contents: read`, set `persist-credentials: false` so the checkout token isn't left in git config for PR-controlled install hooks and audit code to read, and pinned all three actions to commit SHAs rather than movable tags. Also added `package.json` and the workflow file itself to the path filter — both can change what the audit does without currently triggering it.

## Verification

- `pnpm lint` clean, `pnpm build` exit 0
- 0 duplicate/empty heading ids across 2,053 built pages
- 0 h2/h3 slug shifts
- Audit exits 1 on a new broken link (including a second occurrence of a baselined one) and 0 when clean
- The `audit` workflow passed on the previous push in 36s

blockiosaurus (approved):

PR #561, opened 2026-09-16
docs: add Digital Energy to RPC providers (without DAS)

Digitalenergyllc:
Adds Digital Energy to the alphabetical "RPCs without DAS Support" list, between Chainstack and Figment.

Digital Energy is a Solana validator operator running its own RPC infrastructure on dedicated hardware.

- Free tier, no card required.
- Every supported JSON-RPC and WebSocket method documented with runnable request examples and expected responses at portal.digital-energy.io/methods.
- Public status page at status.digital-energy.io.

Non-archive node.

Note: this changes the en locale only. The ko, zh and ja copies of this page carry the same list — happy to mirror the entry into them if that is the convention here.


Digitalenergyllc:
Mirrored into the ja, ko and zh locale files as suggested. Same line, same position between Chainstack and Figment.

